PR #49, opened 2016-04-28
Fixed the Safety indicator for the UHF and Autopilot On/Off switches

chrishajduk84:


chrishajduk84:
This code was used during competition....I'm gonna merge it before it gets lost.

PR #52, opened 2016-07-01
Multigain commands

iehink:
Added setGain command functionality


Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -911,6 +919,10 @@ void readDatalink(void){
             case SET_KIVALUES:
                 setKValues(GAIN_KI,(float*)(&cmd->data));
                 break;
+            case SET_GAINS:
+                int type = (int*) (&cmd->data);
```
CoatedMoose: I don't think this is correct.

Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -515,7 +515,15 @@ void setKValues(int type,float* values){
     for(i=0; i<7; i++){
        setGain(Kchannel[i],type,values[i]);
     }
-};
+}
+
+void setGains(int* type, float* values){
+    int gains[3] = {GAIN_KD,GAIN_KP,GAIN_KI};
```
CoatedMoose: Set this as a global constant so you aren't reallocating every time it is called.

sergei1152: I dont think this is worth a for loop. We'll only ever have 3 types of gains. You can just do: 
```
setGain(type, GAIN_KD, values[0]);
setGain(type, GAIN_KP, values[1]);
setGain(type, GAIN_KI, values[2]);
```
Just add a comment on the function stating that values has to be an array of size 3, with the first value being KD, then KP, then KI to make it clear.

Autopilot/AttitudeManager/AttitudeManager.c
sergei1152: Yeah it doesn't look right. Has this code been tested? Look at the NEW_WAYPOINT case if you want to see how multi part commands (commands with multiple values) are interpreted

iehink: Sorry, when I created the merge request my push had failed and I didn't realize. I will push and create another merge request.

CoatedMoose: You can just push to the same branch, the pork request will update.

iehink: It turns out I don't have permission to push to the branch.

CoatedMoose: Ah, interesting. You can always fork and open a new PR. @scholvat for permission issues.

Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -920,9 +919,11 @@ void readDatalink(void){
                 setKValues(GAIN_KI,(float*)(&cmd->data));
                 break;
             case SET_GAINS:
-                int type = (int*) (&cmd->data);
-                setGains(type,(float*)(&cmd->data));
+            {
+                char* channel = (char*) (&cmd->data);
+                setGains(*channel,(float*)(&cmd->data));
```
CoatedMoose: I believe this is now functionally correct, however it is only correct because the method calls with a built in index offset. This means of you ever call from another location, it is likely to fail in exciting ways. I recommend doing the offset correction on this line and zero indexing in `setGains`. Alternatively, you could make a dedicated struct that describes the command format, and change the methods signatures as appropriate.

CoatedMoose:
LGTM

PR #55, opened 2016-07-20
Gps error check

scholvat:
Removed comp hack and replaced it with a less hacky error checking method.


chrishajduk84:
Simple function. Straightforward. Looks good to me.

sergei1152:
Lets keep in mind that this is a hotfix, not a solution, since we still dont know the underlying problem causing this.

chrishajduk84:
Indeed.

PR #56, opened 2016-07-20
Battery sensors

scholvat:
Added picPilot battery sensor. This PR doesn't include external battery sensor, that will come in a future PR.


Autopilot/AttitudeManager/nbproject/configurations.xml
```diff
@@ -193,8 +196,9 @@
         <property key="use-cci" value="false"/>
         <property key="use-iar" value="false"/>
       </C30>
-      <C30-AS>
-      </C30-AS>
+      <C30-AR>
+        <property key="additional-options-chop-files" value="false"/>
```
sergei1152: Can we not commit these config files? Maybe add the project config folder to the git ignore for now. They still need to be within the repo but theres no pointing in updating them any further.

Autopilot/Path Manager/MainBatterySensor.c
```diff
@@ -6,40 +6,44 @@
...
-//    return time;
-//}
-
-int getBatteryLevel(){
-    percent = currentVoltage;
-    return percent;
+char getMainBatteryLevel(){
+    //return percent of 2 cell battery. 
+    //To calculate percent, the formula takes into account the voltage divider, bit size of a char, minimum 6V of a LiPo 2S, and max charge of 8.4V
+    percent = (((double)(batteryVoltage)/3*13/128)-36*CELLS)*100/(6*CELLS);
```
sergei1152: the voltage to percentage conversion can be performed in the groundstation, and should be. This will keep the picpilot running faster, and it makes it more configurable since within the groundstation you can change the cell count without recompiling and reflashing the pic.

Autopilot/Path Manager/MainBatterySensor.c
```diff
@@ -6,40 +6,44 @@
  */
 #include "main.h"
 #include "../Common/Common.h"
-#include "voltageSensor.h"
+#include "MainBatterySensor.h"
 
 
-char percent = 0;
-int currentVoltage = 0;
+int percent = 0;
```
sergei1152: Theres no reason for this to be an int. A percentage is only ever between 0-100, and a char ranges from -128 to 128.

Autopilot/Path Manager/MainBatterySensor.c
sergei1152: In addition, where did you get this calculation?

Autopilot/AttitudeManager/nbproject/configurations.xml
chrishajduk84: If you add new files they keep them in the project settings,  so you
sometimes need to include them. Otherwise you have to add everything to the
project manually.

Trouble is it also makes changes based on what operating system you are on
and where your working directory is.

It's a problem we had before and kind of just said 'screw it'.

Autopilot/Path Manager/MainBatterySensor.h
```diff
@@ -11,14 +11,15 @@
 #ifdef	__cplusplus
 extern "C" {
 #endif
+    
+#define CELLS 2
```
sergei1152: This should be renamed to something more specific, such as MOTOR_BATTERY_CELL_COUNT or whatever

Autopilot/Path Manager/nbproject/configurations.xml
```diff
@@ -123,6 +127,9 @@
         <property key="use-cci" value="false"/>
         <property key="use-iar" value="false"/>
       </C30>
+      <C30-AR>
+        <property key="additional-options-chop-files" value="false"/>
+      </C30-AR>
```
sergei1152: Same comment about config files as above ^^

Autopilot/AttitudeManager/nbproject/configurations.xml
sergei1152: Yeah I see what your saying. This file specifically has nothing to do with the new file additions though. The other config file does, which is the one that should be commited.

Autopilot/Path Manager/MainBatterySensor.c
scholvat: Changed this to an int so that if the percent is calculated to a value over 128, it would not cause issues.

Autopilot/Path Manager/MainBatterySensor.c
scholvat: Using the PIC documentation, the hardware voltage divider, and the min/max charge of each cell. If you want the raw voltage data coming to the ground station, everything will need be changed to an int, which will take up more space.

scholvat: We are not going to be changing the picpilot battery very often, we always use a 2 cell Lipo, so this calculation won't change much.

sergei1152: its alot more useful for the groundstation to receive a voltage rather than a percentage. The lipo curve isnt linear, and the groundstation can get a better approximation of the battery percentage by using a look up table and such. In addition the groundstation can track the battery voltage over time, and do cool things like provide a fairly accurate battery time left estimate. In addition, the accuracy of the groundstation doesnt matter, but on the picpilot it does. We can modify the percentage to voltage calculation on the groundstation, and use the same source data, in this case the voltage, which remains intact. The same cant be said on the picpilot, since you're losing the original battery voltage.

Autopilot/Path Manager/MainBatterySensor.c
sergei1152: It should never be above 100 in the first place. If it does you've got more issues to worry about

Autopilot/Path Manager/nbproject/configurations.xml
sergei1152: On chris's not, this configuration file is fine, since it adds files to the project

Autopilot/Path Manager/MainBatterySensor.c
sergei1152: Although i see your point. You cant validate that a percentage is within a certain range if it overflows. In any case I dont think you should do the percentage conversion

scholvat:
@sergei1152 I have removed onboard calculations, ground station will need to be updated to handle this data

Autopilot/Common/Common.h
```diff
@@ -92,7 +92,8 @@ typedef struct _PMData { //54 Bytes
     int sp_Altitude; // Meters
     int heading;  //Degrees
     int sp_Heading; //Degrees
-    int batteryLevel;
+    int batteryLevel1;
+    int batteryLevel2;
```
sergei1152: What kind of value will this be? What ranges?

Autopilot/Path Manager/MainBatterySensor.c
```diff
@@ -6,40 +6,34 @@
...
-int getBatteryLevel(){
-    percent = currentVoltage;
-    return percent;
+int getMainBatteryLevel(){
+//return raw voltage value
+    return batteryVoltage;
 }
 
-void initVoltageADC(){
+void initMainBatteryADC(){
```
sergei1152: we're not using this method anywhere, thus theres no need for it to be available publically. Add a static property to this method, so that it can only be used within the same file, and remove it from the MainBattery.h file.

Autopilot/Path Manager/MainBatterySensor.c
```diff
@@ -6,40 +6,34 @@
  */
 #include "main.h"
 #include "../Common/Common.h"
-#include "voltageSensor.h"
+#include "MainBatterySensor.h"
```
sergei1152: Im not sure if this is good naming for the file. What makes this the main battery, and what differentiates from the other battery sensor? Like if Im reading the code, does main battery mean the battery attached to the picpilot powering the components, or the motors? Maybe we can come up with a clearer name here

Autopilot/Path Manager/MainBatterySensor.c
chrishajduk84: I think you got the right idea and I guess it's just my bad code, but I've
been making all my functions public on the entire Picpilot.

It might be something we want to gradually fix.

Autopilot/Common/Common.h
scholvat: I believe it will map the voltage from the AN pin (0-3V3) to a 12 bit integer. So 0V=0, 3V3=111111111111

Autopilot/Path Manager/MainBatterySensor.c
scholvat: I originally named them MainBattery and ExtBattery, but removed external battery because its more complicated. MainBattery is the main power to the PicPilot, and External Battery is the motor batteries (external because it has nothing to do with the picpilot). I can add some comments to clarify this, and it will be more clear when I add the external battery sensor.

sergei1152: Ok, i'm fine with this naming then. Maybe consider naming ExtBattery to MotorBattery to make it really clear when you get to implementing it.

Autopilot/Common/Common.h
sergei1152: Whats 3V3?

Autopilot/Path Manager/MainBatterySensor.c
```diff
@@ -77,3 +65,15 @@ void initMainBatteryADC(){
     AD2CON1bits.ADON = 1;		// Turn on the A/D converter
 
 }
+
+void initMainBatterySensor(){
+    //AN12 is the pin to get the battery information
+    TRISBbits.TRISB12 = 1;
+    initMainBatteryADC();
+
```
sergei1152: get rid of the extra line

Autopilot/Path Manager/MainBatterySensor.c
```diff
@@ -77,3 +65,15 @@ void initMainBatteryADC(){
     AD2CON1bits.ADON = 1;		// Turn on the A/D converter
```
sergei1152: same here

sergei1152:
@scholvat Just updated the PR to remove the extra lines, then feel free to merge 👍

PR #57, opened 2016-09-16
Using channel configuration that matches Steves RC

sergei1152:
- UHF input now uses channel 6 instead of using the flap channel as it did previously
- throttle, left tail, right tail, and roll inputs and outputs match Steve's Controller now, so hookups should be easier

New configuration
Throttle - Channel 1
Roll - Channel 2
Left Tail - Channel 3
Right Tail - Channel 4
Flaps - Channel 5
UHF - Channel 7

Autopilot On/Off is still channel 8 since that's forced by the hardware on the picpilot, and will be the only thing that doesn't match steve's configuration


sergei1152:
Still requires to be tested

sergei1152:
@chrishajduk84 Please review whenever you get the chance

scholvat:
Docs and pre-flight checklist should updated to match: http://docs.uwarg.com/picpilot/pwm-io/
Looks good, meant to do this last term but didn't get around to it.

sergei1152:
Ill make a follow up task and update the docs when this is merged in

sergei1152:
Ok I'll merge this one in. I'll try to test it this weekend, as well as update the documentation

PR #60, opened 2016-09-20
Added new command, TRIGGER_CAMERA_AND_DROP_PROBE, to trigger camera and drop probe.

brenton311:
Embedded Software Bootcamp by Brenton Sutherland.


sergei1152:
Looks good. A minor note is that your command should not be in the 100+ index range, since thats reserved for multipart commands (commands that take in more than a single value). For example the add waypoint command takes in a latitude, longitude, altitude, etc..

However in your case and for the purpose of the bootcamp it would work and is fine.

I'll send you an invite you to asana 👍

PR #64, opened 2016-09-22
Added a way to orbit the plane around a specified location

iehink:
The added lines are the exact same logic as used elsewhere in the code
(copy and pasted into the if statement to keep the order of calculations
the same as before outside the if statement). The followOrbit function
is called here as it simply causes the plane to follow an orbit
independant of an entrance and exit point. To cause the plane to exit
the loop the target waypoint would have to be updated or destroyed. The
size of the orbit can be chaged using the waypoint radius value.

### DIFF
--- a/Autopilot/Path Manager/PathManager.c
+++ b/Autopilot/Path Manager/PathManager.c
@@ -340,7 +340,23 @@ char followWaypoints(PathData* currentWaypoint, float* position, float heading, 
 
             *sp_Heading = (int)followStraightPath((float*)&waypointDirection, (float*)targetCoordinates, (float*)position, heading);
         }
-        else{
+        else{            
+            
+            // if target waypoint is a hold waypoint the plane will follow the orbit until the waypoint is updated or removed
+            if (targetWaypoint->type == HOLD_WAYPOINT) {
+                char turnDirection = waypointDirection[0] * nextWaypointDirection[1] - waypointDirection[1] * nextWaypointDirection[0]>0?1:-1;
+                float euclideanWaypointDirection = sqrt(pow(nextWaypointDirection[0] - waypointDirection[0],2) + pow(nextWaypointDirection[1] - waypointDirection[1],2) + pow(nextWaypointDirection[2] - waypointDirection[2],2)) * ((nextWaypointDirection[0] - waypointDirection[0]) < 0?-1:1) * ((nextWaypointDirection[1] - waypointDirection[1]) < 0?-1:1) * ((nextWaypointDirection[2] - waypointDirection[2]) < 0?-1:1);
+                
+                float turnCenter[3];
+                turnCenter[0] = targetCoordinates[0] + (targetWaypoint->radius/tan(turningAngle/2) * (nextWaypointDirection[0] - waypointDirection[0])/euclideanWaypointDirection);
+                turnCenter[1] = targetCoordinates[1] + (targetWaypoint->radius/tan(turningAngle/2) * (nextWaypointDirection[1] - waypointDirection[1])/euclideanWaypointDirection);
+                turnCenter[2] = targetCoordinates[2] + (targetWaypoint->radius/tan(turningAngle/2) * (nextWaypointDirection[2] - waypointDirection[2])/euclideanWaypointDirection);
+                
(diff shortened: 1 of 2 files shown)


chrishajduk84:
This will be really really useful...
Need to test this ASAP.

Autopilot/Path Manager/PathManager.c
```diff
@@ -340,7 +340,23 @@ char followWaypoints(PathData* currentWaypoint, float* position, float heading,
 
             *sp_Heading = (int)followStraightPath((float*)&waypointDirection, (float*)targetCoordinates, (float*)position, heading);
         }
-        else{
+        else{
```
chrishajduk84: On the right track. So if you look at the logic and compare it to the history of the file: https://github.com/UWARG/PICpilot/blob/54101f9019d5a16ccbd7ea7f513a9ab8b36f3a31/Autopilot/Path%20Manager/PathManager.c

You will notice a problem: This else statement gets called at every turn the aircraft makes. You now put a if statement around what was previously there. In consequence, the plane will never ever get past the first turn/waypoint it makes.

The way the autopilot works is that it alternates between PATH and ORBIT. This "else" statement is the ORBIT.

What you probably need to do is create an if, else statement for this section of code.

Autopilot/Path Manager/PathManager.c
```diff
@@ -340,7 +340,23 @@ char followWaypoints(PathData* currentWaypoint, float* position, float heading,
 
             *sp_Heading = (int)followStraightPath((float*)&waypointDirection, (float*)targetCoordinates, (float*)position, heading);
         }
-        else{
+        else{            
+            
+            // if target waypoint is a hold waypoint the plane will follow the orbit until the waypoint is updated or removed
+            if (targetWaypoint->type == HOLD_WAYPOINT) {
+                char turnDirection = waypointDirection[0] * nextWaypointDirection[1] - waypointDirection[1] * nextWaypointDirection[0]>0?1:-1;
```
scholvat: This is all repeated code from later lines... It would probably be cleaner if you reused that code, possibly put it in a function, or put the halfPlane stuff in an if statement, in case we have to change this in the future.

Autopilot/Path Manager/PathManager.h
```diff
@@ -28,6 +28,7 @@
 //Waypoint types
 #define DEFAULT_WAYPOINT 0
 #define PROBE_DROP_WAYPOINT 1
+#define HOLD_WAYPOINT 2
```
scholvat: How would we send a hold waypoint to the plane? Is the code written on the ground station to be able to add a hold waypoint?

Autopilot/Path Manager/PathManager.c
```diff
@@ -340,7 +340,23 @@ char followWaypoints(PathData* currentWaypoint, float* position, float heading,
 
             *sp_Heading = (int)followStraightPath((float*)&waypointDirection, (float*)targetCoordinates, (float*)position, heading);
         }
-        else{
+        else{            
+            
+            // if target waypoint is a hold waypoint the plane will follow the orbit until the waypoint is updated or removed
+            if (targetWaypoint->type == HOLD_WAYPOINT) {
```
scholvat: Have you thought about how to exit the hold? I would think the best way to do it is to have an exitHold command, that would move on to the next waypoint if the plane was in a hold.

iehink: I will add that in.

Autopilot/Path Manager/PathManager.c
iehink: I'm pretty sure I nested the added if statement into the else, not vice versa. This shouldn't change the current behavior. I could be wrong though.

Autopilot/Path Manager/PathManager.h
iehink: I haven't added anything there. I will after cleaning up this code though.

Autopilot/Path Manager/PathManager.c
scholvat: I'm with you @iehink , not following your logic @chrishajduk84. It should still work normally if there are no hold waypoints.

chrishajduk84: @iehink My bad.  I just took another look and see that you did it correctly. Apologies, I must have been mentally handicapped at that moment.

PR #68, opened 2016-09-29
Hold waypoint

iehink:
Added logic to simply orbit a waypoint if the target waypoint is of type Hold_waypoint. Added method orbitWaypoint which simply calls followOrbit. Added command exit_hold_orbit which will cause the plane to stop orbiting the given waypoint.


Autopilot/Path Manager/PathManager.c
```diff
@@ -48,6 +48,7 @@ int lastKnownHeadingHome = 10;
 char returnHome = 0;
 char doProbeDrop = 0;
 char followPath = 0;
+char inHold = FALSE;
```
sergei1152: For consistency sake lets set this to 0

PR #70, opened 2016-10-01
Doxygen Support for Automatic Generation of Docs

sergei1152:
From now on I'd like the code for the picpilot to be inline-documented via the special `/** **/` tags. Then Doxygen can automatically generate the API docs and upload it to http://uwarg.github.io/PICpilot/ . To do this run the `upload-docs.sh` script


chrishajduk84:
I've never done this.... But it looks like you know what you are doing.

PR #71, opened 2016-10-01
Added ceedling unit testing framework to attitude manager

sergei1152:
Adds support for unit testing the PicPilot using Ceedling following this tutorial http://dmitryfrank.com/articles/unit_testing_embedded_c_applications

Only added on the attitude manager so far. We'll see how it goes. Essentially theres a test folder within the Attitude Manager that will contain all the unit tests for the pic. Ceedling is a unit test runner that contains a mocking library, and a c unit testing framework. They can be run by installing ruby, rake, and running `gem install ceedling` to install ceedling. Then by  running `rake test:all` all the unit tests defined in the tests folder will be run. This will be useful for testing non-hardware related code components, such as things that perform math operations.

The commit is large but I think it's necessary to include all the files.


chrishajduk84:
At the moment this is pretty useless....Correct me if I am wrong, but it does it build yet?

chrishajduk84:
You also seem to be including a library into the repository, which is generally bad practice.... 

Would it be better if we had Installation instructions to run tests? How do the tests run? 
Can you include a little documentation on docs.uwarg.com?

sergei1152:
@chrishajduk84 Yes it builds. I looked into it and there is infact a way to regenerate the library files, so I removed them from the PR. As far as documentation goes, I'm still not familiar enough with it to write any documenation yet. As I start using it more I"ll write a tutorial on how to unit test embedded code on the pic

sergei1152:
But if you'd like to run it yourself, cd into the AttitudeManager test folder, make sure to install ruby and rake, and run `rake test:all`. I put in a single template test file, so there should be 1 test that passes. Eventually we can get a continuous integration server going running tests for us, same as the groundstation and computer vision.

PR #72, opened 2016-10-01
Commented and Refactored UART2.h and UART2.c to remove unused functions

sergei1152:
Cleaned up and documented UART2.c and UART2.h and removed instances of PATH_MANAGER and GPS_OLD from the files, as well as removed `UART2_SendChar` and `UART2_SendString` functions since they werent being used anywhere

### DIFF
--- a/Autopilot/AttitudeManager/UART2.c
+++ b/Autopilot/AttitudeManager/UART2.c
@@ -1,9 +1,14 @@
+/**
+ * @file UART2.c
+ * @author Mitch Hatfield
+ * @date Jun 2013
+ * @brief Contains implementation for initializing UART2 
+ * @copyright Waterloo Aerial Robotics Group 2016 \n
+ *   https://raw.githubusercontent.com/UWARG/PICpilot/master/LICENCE 
+ */
+
 #include "main.h"
 #include "UART2.h"
-
-#if PATH_MANAGER && !GPS_OLD
-UART_RX_Buffer _buff;
-#endif
 
 void InitUART2()
 {
@@ -61,22 +66,14 @@ void InitUART2()
 
 //	IPC7 = 0x4400;	// Mid Range Interrupt Priority level, no urgent reason
 
-        IPC7 = 0x4400;	// Mid Range Interrupt Priority level, no urgent reason
+  IPC7 = 0x4400;	// Mid Range Interrupt Priority level, no urgent reason
 
 	IFS1bits.U2TXIF = 0;	// Clear the Transmit Interrupt Flag
 	IEC1bits.U2TXIE = 0;	// Enable Transmit Interrupts
 	IFS1bits.U2RXIF = 0;	// Clear the Recieve Interrupt Flag
 
-#if PATH_MANAGER && !GPS_OLD
-	IEC1bits.U2RXIE = 1;	// Enable Recieve Interrupts
-#else
-        IEC1bits.U2RXIE = 0;	// Enable Recieve Interrupts
-#endif
+  IEC1bits.U2RXIE = 0;	// Enable Recieve Interrupts
 
-#if PATH_MANAGER && !GPS_OLD
-	U1MODEbits.UARTEN = 1;	// And turn the peripheral on
-	U1STAbits.UTXEN = 1;
-#endif
 	U2MODEbits.UARTEN = 1;	// And turn the peripheral on
 
 	U2STAbits.UTXEN = 1;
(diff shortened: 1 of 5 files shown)


Autopilot/AttitudeManager/UART2.c
```diff
@@ -61,22 +66,14 @@ void InitUART2()
...
+  IPC7 = 0x4400;	// Mid Range Interrupt Priority level, no urgent reason
 
 	IFS1bits.U2TXIF = 0;	// Clear the Transmit Interrupt Flag
 	IEC1bits.U2TXIE = 0;	// Enable Transmit Interrupts
 	IFS1bits.U2RXIF = 0;	// Clear the Recieve Interrupt Flag
 
-#if PATH_MANAGER && !GPS_OLD
-	IEC1bits.U2RXIE = 1;	// Enable Recieve Interrupts
-#else
-        IEC1bits.U2RXIE = 0;	// Enable Recieve Interrupts
```
chrishajduk84: Double check that this doesn't break anything. The else statement could be troublesome. 

If I'm not mistaken UART 2 is used for the debug cable....So we don't need a receive interrupt. Double check to be sure.

Autopilot/AttitudeManager/main.h
```diff
@@ -23,6 +23,7 @@
 //                  heading for the aircraft to follow.
 //  Attitude Manager - Communicates with a IMU (kinematics sensor) in order to
 //                     the desires Pitch, Roll, Yaw on the aircraft.
+//TODO References to these should be stripped out, as they are now in seperate code bases - Serge Sept 2016
```
chrishajduk84: MPLab has a find in projects search feature.

Autopilot/AttitudeManager/UART2.h
```diff
@@ -1,35 +1,18 @@
...
+ * @author Mitch Hatfield
+ * @date Jun 2013
+ * @brief Settings for UART2 communication used for telemetry link 
+ * @copyright Waterloo Aerial Robotics Group 2016 \n
+ *   https://raw.githubusercontent.com/UWARG/PICpilot/master/LICENCE 
  */
 
-#ifndef UART2_H
-#define	UART2_H
+#pragma once
```
chrishajduk84: Does this actually work?

Autopilot/AttitudeManager/UART2.h
```diff
@@ -1,35 +1,18 @@
...
+
 #include "main.h"
 
+/** Initializes the UART2 connection **/
 void InitUART2();
-void UART2_SendString(char *s);
-void UART2_SendChar(char data);
```
chrishajduk84: This will break it. The send string and send char are still used in other parts of the code.

sergei1152: I did a find usage with the IDE and the only time its used is in the implementation

Autopilot/AttitudeManager/UART2.h
sergei1152: It compiled and built the project so I'm assuming so?

Autopilot/AttitudeManager/main.h
sergei1152: Yes, I put that into another task however

Autopilot/AttitudeManager/UART2.c
sergei1152: UART2 is used for the Xbee as documented here http://docs.uwarg.com/picpilot/uart/

chrishajduk84: Okay... But if we disable receive interrupts,  won't that break it?

Autopilot/AttitudeManager/UART2.h
chrishajduk84: Double check debug.c.

When you enable Debugging,  it should use these functions to spit stuff out..... . Actually that's UART1. You are right. I think this is okay.

Autopilot/AttitudeManager/UART2.c
sergei1152: I'm not sure, however the code hasn't changed from what was there before. The preprocessor would have just pasted in `IEC1bits.U2RXIE = 0;`, so this change should affect any of the exisitng functionality

chrishajduk84: If you are removing IEC1bits.U2RXIE = 0; it won't paste that in anymore though...

sergei1152: I havent removed it. I moved it to be outside the if else. Check the file diff

PR #78, opened 2017-01-18
BootCamp by Tahir Dawson

t2dawson:


sergei1152:
PROBE_DROP_TRIGGER isn't defined anywhere,  so your function is never going to execute. I'd actually expect the preprocessor to give you a compilation error. 

Anyways, you need to define this command in the proper place for the code to work.

sergei1152:
Since you also made a separate. .h. File containing your function, your code definitely won't compile since you're missing an include. Make sure your code compiles in MPLabX.

t2dawson:
Hi,

Is the function otherwise okay or should there be any changes? Also i
defined PROBE_DROP_TRIGGER in the commands file where the other commands
were defined. I will check to make sure it's possible i didnt push the
change

Regards,
Tahir Dawson

t2dawson:
Okay sorry for that i forgot to push the command.h file. I will also make sure it compiles right and update it.

t2dawson:
I have added in the missing header and the command. Tried it on MPLabX and it does compile. Please let me know if there are any other changes to make here

t2dawson:
Hi,

My email is t2dawson@uwaterloo.ca. I have already joined on slack. Will
definitely come by today and work on the task.

PR #80, opened 2017-01-26
Update of PR # 74 (Update Path Manager GPS lock)

t2dawson:


chrishajduk84:
Just to verify is this a legitimate commit? Or is this just for the boot camp?

Autopilot/Path Manager/PathManager.c
```diff
@@ -667,12 +667,37 @@ void copyGPSData(){
     pmData.checkbyteDMA2 = generatePMDataDMAChecksum2();
 }
 
+// check if power on reset, brown out reset, watchdog time out or software reset flags are set
+
+ char checkForReset(){
+     if (RCONbits.POR || RCONbits.BOR || RCONbits.WDTO || RCONbits.SWR || RCONbits.EXTR){
+         // clear RCON(Reset control) bits)
```
chrishajduk84: This function might not work.  Or it may break the functionality of StartupErrorCodes.c. When you check these special function registers,  they automatically reset,  as such they can be only read once. 

StartupErrorCodes.c contains get() functions to retrieve this information. Use that instead.

Autopilot/Path Manager/PathManager.c
```diff
@@ -667,12 +667,37 @@ void copyGPSData(){
     pmData.checkbyteDMA2 = generatePMDataDMAChecksum2();
 }
 
+// check if power on reset, brown out reset, watchdog time out or software reset flags are set
+
+ char checkForReset(){
```
scholvat: Either rename the function to something more obvious, or put a comment saying what the return value represents

Autopilot/Path Manager/PathManager.c
```diff
@@ -667,12 +667,37 @@ void copyGPSData(){
...
+      home.altitude = gpsData.altitude;
+     }
+  }
 //returns 1 if gps location makes sense, 0 if not
 char gpsErrorCheck(double lat, double lon){
     if(abs(lon - RELATIVE_LONGITUDE)<GPS_ERROR && abs(lat - RELATIVE_LATITUDE)<GPS_ERROR){
         return 1;
     }
+    else{
+        GPSLock();
```
scholvat: I don't think you should call GPSLock() from this method, this method is only meant to check for error so it shouldn't perform any other function. Possibly put it around line 648

Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -21,6 +21,7 @@
 #include "ProgramStatus.h"
 #include <string.h>
 
+
```
sergei1152: Get rid of the commit that causes the file changes here, since nothing changed in these files.

Autopilot/AttitudeManager/AttitudeManager.h
```diff
@@ -15,6 +15,7 @@
 #include "net.h"
 #include "../Common/Common.h"
 
+
```
sergei1152: Same thing, nothing changed in these files, so get rid of the commit / get rid of the space.

Autopilot/Path Manager/PathManager.h
```diff
@@ -58,6 +58,8 @@ void copyGPSData(void);
 char gpsErrorCheck(double lat, double lon);
 void checkAMData(void);
 float getWaypointChecksum(void);
+void GPSLock();
+char checkForReset();
```
sergei1152: No need for either of these functions to be declared inside the header, since they are only internally used within the Path Manager. These functions should be declared static so that they only exist within the scope of the PathManager.cpp file, and removed from the header.

Autopilot/Path Manager/PathManager.c
```diff
@@ -667,12 +667,37 @@ void copyGPSData(){
     pmData.checkbyteDMA2 = generatePMDataDMAChecksum2();
 }
 
+// check if power on reset, brown out reset, watchdog time out or software reset flags are set
+
+ char checkForReset(){
+     if (RCONbits.POR || RCONbits.BOR || RCONbits.WDTO || RCONbits.SWR || RCONbits.EXTR){
+         // clear RCON(Reset control) bits)
+         RCONbits.BOR = 0;
```
sergei1152: There should be no need to reset the startup error codes. As chris stated you need to make sure that they are set to the same value that they were at before.

Autopilot/Path Manager/PathManager.c
```diff
@@ -667,12 +667,37 @@ void copyGPSData(){
...
+         RCONbits.POR = 0;
+         RCONbits.SWR = 0;
+         RCONbits.WDTO = 0;
+         return 1;
+     }
+     else 
+         return 0;
+ } 
+ void GPSLock(){
+     if(checkForReset()){
```
sergei1152: No need for the checkForReset function at all, since its only ever going to be used once. I'd just call this function `checkFirstGPSLock` or something, check the startup error code registers, and then set the gps coordinates.

Autopilot/Path Manager/PathManager.c
```diff
@@ -667,12 +667,37 @@ void copyGPSData(){
     pmData.checkbyteDMA2 = generatePMDataDMAChecksum2();
 }
 
+// check if power on reset, brown out reset, watchdog time out or software reset flags are set
+
+ char checkForReset(){
+     if (RCONbits.POR || RCONbits.BOR || RCONbits.WDTO || RCONbits.SWR || RCONbits.EXTR){
```
sergei1152: Comment on the registers you're checking. Remember that if a watchdog reset or software reset occured, we DO NOT want to reset the home coordinates. The only time we'll want it reset is if it was due to a power or hardware reset.

sergei1152:
Also do a git squash (done with a rebase, you can look it up) on the commits. Your bootcamp commits shouldnt be in this PR.

Autopilot/Path Manager/PathManager.c
sergei1152: I see why you're resetting the startup codes. Instead add a flag in the code (a boolean) and in line 654 and 655, the first time you receive one or more satellite fixes, set the flag and run the reset home coordinates function.

chrishajduk84: Point I was trying to make earlier... By resetting the flags here,  it is breaking functionality elsewhere.  Simply calling getErrorFlags() /checkErrorFlags()  would be a better way to clear/get the flags.

PR #81, opened 2017-01-29
Fixed data corruption issues with multipart waypoint commands

ianjfrosst:


Autopilot/Common/Common.h
```diff
@@ -57,15 +57,28 @@
...
+ char           : 1 byte
+ int            : 2 bytes
+ long int       : 4 bytes
+ float          : 4 bytes
+ long double    : 8 bytes
+ */
+
+
+/*  WAYPOINT WRAPPER IS USED FOR NETWORKING 
+    DO NOT CHANGE THIS UNLESS YOU KNOW WHAT YOU ARE DOING */
```
scholvat: All multipart command structs should be grouped together in the same file, but that is out of the scope of this PR

PR #83, opened 2017-02-01
Update PathManager.c with home coordinates locking conditions

t2dawson:
Closed down the previous request due to issues with git when trying to remove old (bootcamp) commits. please take a look at this and let me know if any changes have to be made.

### DIFF
--- a/Autopilot/Path Manager/PathManager.c
+++ b/Autopilot/Path Manager/PathManager.c
@@ -13,6 +13,7 @@
 #include "MainBatterySensor.h"
 #include "airspeedSensor.h"
 #include "ProbeDrop.h"
+#include "StartupErrorCodes.h"
 
 #include "InterchipDMA.h"
 
@@ -21,6 +22,11 @@
 #include <stdlib.h>
 #include "../Common/UART1.h"
 #endif
+
+#define BOR_RESET (1<<1)
+#define POR_RESET 1
+
+static void checkForFirstGPSLock(); // function prototype for static function at line 679
 
 extern GPSData gpsData;
 extern PMData pmData;
@@ -653,6 +659,7 @@ void copyGPSData(){
         pmData.speed = gpsData.speed;
         pmData.satellites = (char)gpsData.satellites;
         pmData.positionFix = (char)gpsData.positionFix;
+        checkForFirstGPSLock();
     }
     pmData.batteryLevel1 = getMainBatteryLevel();
     pmData.batteryLevel2 = 5;
@@ -665,6 +672,19 @@ void copyGPSData(){
     pmData.pathFollowing = followPath;
     pmData.checkbyteDMA1 = generatePMDataDMAChecksum1();
     pmData.checkbyteDMA2 = generatePMDataDMAChecksum2();
+}
+
+static void checkForFirstGPSLock(){
+	static char gpsLockFlag = 1;
+	
+	if(((getErrorCodes() & BOR_RESET) || (getErrorCodes() & POR_RESET)) && gpsLockFlag){
+		home.latitude = gpsData.latitude;
+		home.longitude = gpsData.longitude;
+		home.altitude = gpsData.altitude;
+			
+		gpsLockFlag = 0;
+	}
+	
 }
 
 //returns 1 if gps location makes sense, 0 if not
--- a/Autopilot/Path Manager/PathManager.c
+++ b/Autopilot/Path Manager/PathManager.c
@@ -23,8 +23,6 @@
 #include "../Common/UART1.h"
 #endif
 
-#define BOR_RESET (1<<1)
-#define POR_RESET 1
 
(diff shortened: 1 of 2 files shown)


Autopilot/Path Manager/PathManager.c
```diff
@@ -22,6 +23,11 @@
 #include "../Common/UART1.h"
 #endif
 
+#define BOR_RESET (1<<1)
```
sergei1152: It makes more sense to define these in StartUpErrorCodes.h, so that other things may use them. Go with a verbose name like `STARTUP_ERROR_BROWN_OUT_RESET` so that its more descriptive

Autopilot/Path Manager/PathManager.c
```diff
@@ -667,6 +674,19 @@ void copyGPSData(){
     pmData.checkbyteDMA2 = generatePMDataDMAChecksum2();
 }
 
+static void checkForFirstGPSLock(){
+	static char gpsLockFlag = 1;
+	
+	if(((getErrorCodes() & BOR_RESET) || (getErrorCodes() & POR_RESET)) && gpsLockFlag){
```
sergei1152: Theres a bit of overhead in calling a function. its not a big deal, but i'd suggest making another character variable to store the result of error codes, then checking that.

Autopilot/Path Manager/PathManager.c
```diff
@@ -667,6 +674,19 @@ void copyGPSData(){
     pmData.checkbyteDMA2 = generatePMDataDMAChecksum2();
 }
 
+static void checkForFirstGPSLock(){
+	static char gpsLockFlag = 1;
```
sergei1152: This isn't going to work since you're resetting the flag every time the function is called. If you want it to be persistent you'll have to call this outside the function scope.

t2dawson: shouldn't the variable retain it's value since it's static?

Autopilot/Path Manager/PathManager.c
t2dawson: It'd be great if you can guide me a little here. I cant use a static variable since the value of the errors is determined at runtime and when using a non-static one, the getErrorCodes() function would be called everytime the checkForFirstGPSLock() function is executed (which is the same as now except function calls would be cut by half). The other way I can think of is using the lastRuntimeErrors variable itself from the StartUpErrorCodes.c since it is defined as a global variable there so i can extern it and use that.

Autopilot/Path Manager/PathManager.c
sergei1152: In this case it wont, since you're resetting it at the begining of the function. It would retain its value, if say, you only declared it at the top (not set it), and then set it through some form of if conditions, which wouldnt work in this case.

Autopilot/Path Manager/PathManager.c
sergei1152: I would just move your `gpsLockFlag` variable outside the function, which will take care of the static problem. Calling the function twice isnt a very big deal, so you can keep that. Otherwise you can just use a nested if:
```c
if (gpsLockFlag){
    char error_codes = getErrorCodes();
    if (error_codes & BOR_RESET || error_codes & POR_RESET){
     //your logic
    }
}
```

```

Comment (sergei1152) [inline Autopilot/Path Manager/PathManager.c]: Also the way you placed your if statement, getErrorCodes() will always be called twice, no mater whether gpsLockFlag is 1 or 0. The arrangement of variables in an if do matter. So in your case:
```c```
if(((getErrorCodes() & BOR_RESET) || (getErrorCodes() & POR_RESET)) && gpsLockFlag)
```
The first statement will always be executed (the 2 getErrorCodes() calls), even if gpsLockFlag is 0.

However in this case:
```c
if(gpsLockFlag && ((getErrorCodes() & BOR_RESET) || (getErrorCodes() & POR_RESET)))
```
If gpsLockFlag is 0, neither of the functions will be called.

http://stackoverflow.com/questions/1331318/how-an-if-a-b-statement-is-evaluated

Autopilot/Path Manager/PathManager.c
t2dawson: So i actually looked it up and according to http://stackoverflow.com/questions/5033627/static-variable-inside-of-a-function-in-c it should still retain it's value even if declared inside the function since the initialization is done at compile time only. I also then tried it with a small piece of test code to confirm and it does retain the value. I can talk to you more about this in the bay today

sergei1152: ah ok 👍  I haven't used static variables inside functions much so I wouldn't have known. That should work then

t2dawson: Alright I'll push the changes right now then

t2dawson: Changes pushed

PR #84, opened 2017-02-04
InputCompare module refactor, documentation, and bug fixes

sergei1152:
**Changes**:
- Renamed some of the variable names and put in more documentation describing the input compare module (used for PWM)
- Moved initTimer2 function into the timer.h and timer.c modules
- Declared necessary variables in InputCompare.c and .h as static if applicable
- Fixed bug where intermittently PWM values would read extremely low (~-31000) and extremely high (~31000). This was because we were using a signed integer variable to read the timer2 values to be used for comparisons (should be an unsigned integer)
- Fixed issue with hot swapping PWM cables by clearing out the input compare buffer after every read

This branch was thoroughly tested using an RC receiver and all 8 channels. Consistent PWM values were received for all 8 channels

### DIFF
--- a/Autopilot/AttitudeManager/AttitudeManager.h
+++ b/Autopilot/AttitudeManager/AttitudeManager.h
@@ -83,8 +83,11 @@
 #define IMU_PITCH_RATE   1
 #define IMU_ROLL_RATE    0
 
-
-//channels
+/**
+ * RC receiver channel configuration. Note that
+ * Note that only channel 8 will work hardware-wise as the
+ * autopilot channel, as it directly controls the relays
+ */
 #define THROTTLE_IN_CHANNEL 1
 #define THROTTLE_OUT_CHANNEL 1
 #define ROLL_IN_CHANNEL 2
@@ -96,7 +99,7 @@
 #define FLAP_IN_CHANNEL 5
 #define FLAP_OUT_CHANNEL 5
 #define UHF_STATUS_IN_CHANNEL 7
-#define PROBE1_OUT_CHANNEL 6
+#define PROBE1_OUT_CHANNEL 6 //TODO: Remove the probe drop channels as they are no longer used
 #define PROBE2_OUT_CHANNEL 7
 #define PROBE3_OUT_CHANNEL 8
 #define AUTOPILOT_ACTIVE_IN_CHANNEL 8 //Do not change this value, as only 8 will work hardware-wise
--- a/Autopilot/AttitudeManager/InputCapture.c
+++ b/Autopilot/AttitudeManager/InputCapture.c
@@ -9,18 +9,18 @@
 #include "timer.h"
 
 /**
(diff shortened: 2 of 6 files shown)


Autopilot/AttitudeManager/AttitudeManager.h
```diff
@@ -96,7 +99,7 @@
 #define FLAP_IN_CHANNEL 5
 #define FLAP_OUT_CHANNEL 5
 #define UHF_STATUS_IN_CHANNEL 7
-#define PROBE1_OUT_CHANNEL 6
+#define PROBE1_OUT_CHANNEL 6 //TODO: Remove the probe drop channels as they are no longer used
```
chrishajduk84: Rename them to AUX_OUT

Autopilot/AttitudeManager/InputCapture.c
```diff
@@ -1,298 +1,318 @@
...
-        IC8CONbits.ICM = 0b00; // Disable Input Capture 8 module
-        IC8CONbits.ICTMR = 1; // Select Timer2 as the IC1 Time base
-        IC8CONbits.ICI = 0b11; // Interrupt on every capture event
-        IC8CONbits.ICM = 0b001; // Generate capture event on every Rising edge
 
-        // Enable Capture Interrupt And Timer2
-        IPC5bits.IC8IP = 7; // Setup IC8 interrupt priority level - Highest
-        IFS1bits.IC8IF = 0; // Clear IC8 Interrupt Status Flag
-        IEC1bits.IC8IE = 1; // Enable IC8 interrupt
+    while (IC7CONbits.ICBNE) {
```
chrishajduk84: What is the purpose of this statement? 
BNE?

sergei1152: So the pic has a FIFO buffer of size 4 that stores a the timer values.  When you read from IC1BUF you're actually retrieving one of the values from the buffer, and removing it (so shrinking it by one). The ICBNE register is 1 if the buffer is not empty (if there are values to read from it). In our case, since we trigger the interrupt on every edge, the size of the buffer should only ever be 1. However I noticed some strange behaviour when I unplugged a PWM cable from the input and plugged it back in. It seemed to have been an issue with the buffer still containing values, thus it was reading the wrong things. Im still not 100% sure, but I know that with this code I didn't encounter those issues.

Autopilot/AttitudeManager/AttitudeManager.h
sergei1152: I'll probably move these typedefs out into PWM.h since it makes more sense. We've got too much defined in AttitudeManager.h. I'll do this in a seperate PR

Autopilot/AttitudeManager/InputCapture.c
chrishajduk84: Cool... Good

PR #85, opened 2017-02-06
Refactored and documented PWM and OutputCompare Modules

sergei1152:
**Changes**:
- Documented OutputCompare and PWM modules
- Moved out Timer2 logic into the timer module (such as setting Timer2 period, etc)
- Moved the TImer2 initialization call to the PWM module (used to be in InputCompare)
- Removed any unused functions from PWM, InputCompare, and OutputCompare modules
- renamed and moved around the define statements to make more sense

Note that this is only a refactor, however hasn't been tested yet. I'll test this afterwords with an integration test

### DIFF
--- a/Autopilot/AttitudeManager/PWM.h
+++ b/Autopilot/AttitudeManager/PWM.h
@@ -20,6 +20,12 @@ extern "C" {
 #define MAX_PWM 1024
 #define MIN_PWM -1024
 
+//TODO: These were defined in OutputCompare.h    
+#define UPPER_PWM 1284//1320//1411//941
+#define LOWER_PWM 642//705//470
+#define MIDDLE_PWM 963//981//(int)((UPPER_PWM - LOWER_PWM)/2) + LOWER_PWM//706
+#define SP_RANGE MAX_PWM//(UPPER_PWM - MIDDLE_PWM)
+
 // Function Prototypes
 /*****************************************************************************
  * Function: void initPWM(char inputChannels, char outputChannels);
--- a/Autopilot/AttitudeManager/PWM.h
+++ b/Autopilot/AttitudeManager/PWM.h
@@ -1,116 +1,56 @@
-/* 
- * File:   PWM.h
- * Author: Chris Hajduk
- *
- * Created on August 21, 2014, 11:45 PM
+/**
+ * @file PWM.h
+ * @author Chris Hajduk
+ * @created August 21, 2014, 11:45 PM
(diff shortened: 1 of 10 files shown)


Autopilot/AttitudeManager/PWM.h
```diff
@@ -1,150 +1,96 @@
...
- *                                  For instance, if the range is from -1024 to 1024,
- *                                  the scale factor would be (half the desired output range)/1024
- *        unsigned int signalOffset -> Compensation for a non-zero offset on the input. This is
- *                                      known as trim.
- *
- * Output: None.
- *
- *****************************************************************************/
-void PWMOutputCalibration(unsigned int channel, float signalScaleFactor, unsigned int signalOffset);
+#define SP_RANGE MAX_PWM//(UPPER_PWM - MIDDLE_PWM)
```
chrishajduk84: Is this still correct?  From what I recall SP_RANGE was indeed half of the actual period.... But  I don't remember fully.

chrishajduk84: Consider changing this variable name,  since you are in the process of Refactoring.

sergei1152: I've actually got no idea what this variable does, hence I didnt document it or change it. Here is where its used:

In orientation control.h:
```
#define THROTTLE_SCALE_FACTOR SP_RANGE///Remove this * 2 if having problems
#define FLAP_SCALE_FACTOR SP_RANGE
```
In cameraManager.h
//Constants
```
#define GIMBAL_PWM_RANGE SP_RANGE //Range one way (IE. Positive or Negative range, not both)
```
And in AttitudeManager:
```
int getPitchAngleInput(char source){
    if (source == PITCH_RC_SOURCE){
        return (int)((input_RC_PitchRate / ((float)SP_RANGE / MAX_PITCH_ANGLE) ));
    }
    else if (source == PITCH_GS_SOURCE){
        return input_GS_Pitch;
    }
    else
        return 0;
}
```

If you could shed some light as to what this does, that'd be good. I can document the variable after

chrishajduk84: Its supposed to just be a scaling factor for half the PWM range. Sort of similar to how you typically convert between deg and rad using 180/pi.

I guess an appropriate definition would be the PWM half range.
Which in terms of readability is better than just "MAX_PWM".

The actual appropriate definition should probably be: (MAX_PWM - MIN_PWM)/2
Which in this case (and probably all cases) is equal to 1024.

PR #87, opened 2017-02-06
New UHF Disconnect detection

sergei1152:
**Changes**:
- Time is now an unsigned integer (instead of signed int) which makes more sense
- New UHF disconnect logic. Instead of using a separate UHF channel and comparing its value between two min and max thresholds, detection is now done by tracking the last time an edge was detected on an arbitrary channel. For this to work the RC receivers must be configured to disable their PWM outputs on a disconnect, thus their fail safe options must be disabled. This was done on the orange RX already, and can be done on the ezUHF as well as most likely the dragon link
- Re-write checkUHFStatus function in AttitudeManager. I noticed a bug in that UHFTimer was always 0 (never set), thus if competition mode was turned on and a uhf disconnect occured, the plane would go into kill mode immediately.

Only 1 channel was chosen (channel 7) for the detection logic since it requires calling an external function inside an ISR, which I wanted to limit. Channel 7 was chosen as this will be the RSSI input channel, which will be configured in a later PR

**Note**
This has not been tested yet. I'll test this along with the other changes in an upcoming workday

### DIFF
--- a/Autopilot/AttitudeManager/InputCapture.c
+++ b/Autopilot/AttitudeManager/InputCapture.c
@@ -24,6 +24,13 @@ static char new_data_available[8];
  */
 static unsigned int capture_value[8];
 
+/**
+ * Used to detect a UHF disconnect. Keeps track of the last time a falling edge
+ * was detected. Currently on the Channel 7 ISR will change this variable, however
+ * this can be configured to be any one of the other channels
+ */
+static unsigned long int last_capture_time;
+
 static void calculateICValue(unsigned char channel);
 
 unsigned int* getICValues()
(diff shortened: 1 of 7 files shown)


Autopilot/AttitudeManager/InputCapture.c
```diff
@@ -285,6 +296,7 @@ void __attribute__((__interrupt__, no_auto_psv)) _IC7Interrupt(void)
     } else {
         end_time[6] = IC7BUF;
         new_data_available[6] = 1;
+        last_capture_time = getTime(); //Capture the time. Used for detecting disconnects
     }
```
chrishajduk84: So the problem with this, is even if your receiver disconnects,  it outputs either the last known settings or it will output the failsafe signal (which can be programmed into the controller). 

This will only work if you physically pull the PWM cable.

sergei1152: As discussed, the receiver is set up so that it outputs 0 on all PWM channels on a disconnect

sergei1152:
I tested this in the lab and it works. Merging this in

PR #90, opened 2017-02-12
Receiver PPM Implementation

sergei1152:
**Changes**:
- Receiver input of up to 16 channels using PPM on channel 1
- debug functions now also output a carriage return (so you dont have to keep reconfiguring realterm or putty)

Tested on both the OrangeRX and ezUHF. Note that the ezUHF has an inverted PPM output, thus an option for it was also added. Both the receivers were configured to output ppm on channel 5/port 5 output. 

Both were also configured to output link quality on channel 6 and rssi on channel 7.

Also note that the ezUHF doesnt have the option to disable its PPM/PWM channels as a fail safe, thus channel detection cant be used for it. The dragon link does, so i figure its not worth implementing an alternative method of doing this, since we'll probably end up using the dragon link on the plane and the orangerx on the quad.

The way I imlemented it, you can either use PPM or you can use PWM, but not both. It is however possible to do so, although the channel mapping will get complicated. Since we have no need for this right now, no need to implement it yet.

Also if switching between PPM or PWM, the input capture ranges have to be recalibrated. From my testing, the pulse-widths with PPM were shorter than PWM (depending on the receiver)

Other links:
https://uwarg-docs.atlassian.net/wiki/display/PicPilot/OrangeRX+433Mhz+OpenLRSng+UHF+System
https://uwarg-docs.atlassian.net/wiki/display/PicPilot/ezUHF+433Mhz+Radio+System

### DIFF
--- a/Autopilot/AttitudeManager/InputCapture.c
+++ b/Autopilot/AttitudeManager/InputCapture.c
@@ -114,9 +114,9 @@ unsigned int* getICValues(unsigned long int sys_time)
  */
 void initIC(unsigned char initIC)
 {
-    //If using PPM, we want to unconditionally turn on channel 1
+    //If using PPM, we want to unconditionally turn on channel 7
     #if USE_PPM
-      initIC = 1;
+      initIC = 0b01000000;
       ppm_index = 0;
     #endif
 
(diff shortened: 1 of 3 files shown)


sergei1152:
yeah i tested it on both receivers. links to oscilloscope results above

sergei1152:
I thought only channel 8 was connected to the safety relay?

sergei1152:
Like the relay chip only has a single input compare pin, and thats connected directly to channel 8. @chrishajduk84

chrishajduk84:
Yes that's correct. Only one PWM pin for the safety is available.

scholvat:
We would want to reserve channels 1-4 for the PWM motor channels coming from the flight controller for the quad since those are the only 4 channels that have the bypass to go straight from input to output

sergei1152:
The flight controller should be able to take in PPM tho

scholvat:
Channel 7 should be fine

sergei1152:
However thinking about this I realized we can have it both ways. If I set the PPM channel to be channel 7, then we can use PPM for the picpilot and still leverage the relay chip without any external hardware. The receiver should be configured to output link quality on channel 7 and rssi on channel 8, and output the PPM signal on port 7 or 8.

With this, the picpilot still uses the PPM input (which is nice for development), however when the relay switches, it'll work as usual as it'll still have all the inputs (nothing should be connected to ppm output).

Then we dont actually need to have any sort of ppm converter, although one would be nice for easier wiring.

PR #91, opened 2017-02-18
Zarak Yahya: Bootcamp

zarak3268:


sergei1152:
If you look in the bootcamp description, you'll see:

The groundstation will send a value along with the command that can either be 1, 2, or 3, indicating the probe to drop.

Currently the way you implemented it you can only drop probe 1.

zarak3268:
Which variable contains the value that determines which probe to drop? The only one that I suspect is the "data" character array under the command struct in the net.h file, but that's not an integer.

ianjfrosst:
In C, `char` is a generic 1-byte data type. Some of the time, this represents an ASCII character value. But in the PICpilot's case, `char` is the same as `uint8_t` (unsigned integer, 8 bits).  So `data` is just an array of bytes.

zarak3268:
Okay, thanks for the clarification. Changed the code a bit.

ianjfrosst:
Looks good to me. @sergei1152?

PR #94, opened 2017-02-23
Fixed PPM implementation

ianjfrosst:


sergei1152:
:+1: FYI this if from when I changed the default PPM channel from 1 to 7. I forgot to change somethings, resulting in the autopilot hanging

PR #95, opened 2017-02-24
Fixed integer underflow on PWM inputs

ianjfrosst:


Autopilot/AttitudeManager/PWM.c
```diff
@@ -71,19 +71,19 @@ void initPWM(unsigned char inputChannels, unsigned char outputChannels)
...
-        
+
         if(!channel_enabled){ //if the channel is disabled
             pwm_inputs[channel] = DISCONNECTED_PWM_VALUE;
         } else if (ic_values[channel] == 0 && channel_enabled) { //if the channel is enabled but disconnected
             pwm_inputs[channel] = DISCONNECTED_PWM_VALUE;
             disconnected_pwm_inputs = disconnected_pwm_inputs | (1 << channel); //set the bit as 1
         } else { //otherwise if its a connected, enabled channel, calculate its value
-            pwm_inputs[channel] = (int) ((ic_values[channel] - input_offsets[channel]) * input_scale_factors[channel]);
+            pwm_inputs[channel] = ((int)(ic_values[channel]) - input_offsets[channel]) * input_scale_factors[channel];
```
sergei1152: The type is already an integer, you shouldn't need the cast here

ianjfrosst: I think the int cast there is the "better" way to do it. My initial change was a quick fix, and introduced compiler warnings.

PR #97, opened 2017-03-02
Refactored vehicle control

ianjfrosst:
I'm so sorry.

Summary of changes:
- Started refactoring vehicle code, to improve modularity.
- Reimplemented PPM better (should have been its own PR. Oops.)
- Fixed timers (was going to be its own PR, but I didn't finish with the branch).
- Removed some useless crufty functions and flag variables.
- Replaced the UNARMED loop with a StateMachine call.
- Updated .gitignore, removed files that should be gone.
- Updated compiler version.

Left to do:
- Rewrite PID and control functions for more flexibility (quads fly differently than planes).
- Check data rates & reliability over Xbee link.
- Ensure flight control is running often enough (timer interrupts).
- Do something else with my free time ?????


Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -385,8 +381,15 @@ void setHeadingSetpoint(int setpoint){
 void inputCapture(){
     int* channelIn;
     channelIn = getPWMArray(getTime());
-    inputMixing(channelIn, &input_RC_RollRate, &input_RC_PitchRate, &input_RC_Throttle, &input_RC_YawRate, &input_RC_Flap);
-
+    
+    inputMixing(channelIn, &input_RC_RollRate, &input_RC_PitchRate, &input_RC_Throttle, &input_RC_YawRate);
+    
+#if VEHICLE_TYPE == FIXED_WING
```
sergei1152: I dont think AttitudeManager should be aware of the vehicle type we're using, so we shouldnt have these defines. In this case, it'd probably be better if inputMixing takes in either a struct, and each vehicle file can choose to either populate or not populate the Flap field.

Autopilot/AttitudeManager/FixedWing.c
```diff
@@ -0,0 +1,248 @@
...
+ * Created on July 2, 2015, 8:04 PM
+ */
+
+#include "PWM.h"
+#include "AttitudeManager.h"
+#include "delay.h"
+#include "FixedWing.h"
+#include "ProgramStatus.h"
+
+#if VEHICLE_TYPE == FIXED_WING
```
sergei1152: I think you should remove this. Instead the files that require it can either choose to include or exclude the FixedWing.h file based on the Vehicle type property

Autopilot/AttitudeManager/FixedWing.c
```diff
@@ -0,0 +1,248 @@
+/*
+ * File:   FixedWing.c
```
sergei1152: Add doxygen style file headers if you're creating a new file. In this case, @file, @author, etc. Look at examples in my PR

Autopilot/AttitudeManager/FixedWing.c
```diff
@@ -0,0 +1,248 @@
...
+#if VEHICLE_TYPE == FIXED_WING
+
+int outputSignal[NUM_CHANNELS];
+int control_Roll, control_Pitch, control_Yaw, control_Throttle;
+
+void initialization(){
+    setPWM(THROTTLE_OUT_CHANNEL, MIN_PWM);
+    
+    int channel = 0;
+    for (; channel < NUM_CHANNELS; channel++) {
```
sergei1152: To keep the for loop style consistent, declare the channel above, and initialize it to 0 inside the for loop.

Autopilot/AttitudeManager/FixedWing.c
```diff
@@ -0,0 +1,248 @@
...
+#include "PWM.h"
+#include "AttitudeManager.h"
+#include "delay.h"
+#include "FixedWing.h"
+#include "ProgramStatus.h"
+
+#if VEHICLE_TYPE == FIXED_WING
+
+int outputSignal[NUM_CHANNELS];
+int control_Roll, control_Pitch, control_Yaw, control_Throttle;
```
sergei1152: if these only exist in the scope of the file declare them as static

Autopilot/AttitudeManager/FixedWing.c
```diff
@@ -0,0 +1,248 @@
...
+    }
+    setProgramStatus(MAIN_EXECUTION);
+}
+
+void takeOff(){
+
+}
+void landing(){
+    
+}
```
sergei1152: remove these functions

Autopilot/AttitudeManager/FixedWing.c
```diff
@@ -0,0 +1,248 @@
...
+
+void inputMixing(int* channelIn, int* rollRate, int* pitchRate, int* throttle, int* yawRate){
+        if (getControlPermission(ROLL_CONTROL_SOURCE, ROLL_RC_SOURCE,ROLL_CONTROL_SOURCE_SHIFT)){
+            (*rollRate) = channelIn[ROLL_IN_CHANNEL - 1];
```
sergei1152: No need for the brackets

Autopilot/AttitudeManager/FixedWing.c
```diff
@@ -0,0 +1,248 @@
...
+        channelOut[ROLL_OUT_CHANNEL - 1] = MAX_ROLL_PWM;
+        // Limits the effects of the integrator, if the output signal is maxed out
+//        if (getIntegralSum(ROLL) * getGain(ROLL, GAIN_KI) > sp_RollRate - sp_ComputedRollRate) {
+//            setIntegralSum(ROLL, getIntegralSum(ROLL)/1.1);
+//        }
+    }
+    if (channelOut[ROLL_OUT_CHANNEL - 1] < MIN_ROLL_PWM) {
+        channelOut[ROLL_OUT_CHANNEL - 1] = MIN_ROLL_PWM;
+        // Limits the effects of the integrator, if the output signal is maxed out
+//        if (getIntegralSum(ROLL) * getGain(ROLL, GAIN_KI) < sp_RollAngle - sp_ComputedRollRate) {
```
sergei1152: get rid of the outdated comments

Autopilot/AttitudeManager/FixedWing.c
```diff
@@ -0,0 +1,248 @@
...
+    
+    //Flaps
+    if (channelOut[FLAP_OUT_CHANNEL - 1] > MAX_PWM){
+        channelOut[FLAP_OUT_CHANNEL - 1] = MAX_PWM;
+    }
+}
+
+void highLevelControl(){
+    //If the commands come from the ground station
+    if (getControlPermission(ALTITUDE_CONTROL,ALTITUDE_CONTROL_ON,ALTITUDE_CONTROL_SHIFT) && getControlPermission(ALTITUDE_CONTROL_SOURCE,ALTITUDE_GS_SOURCE,ALTITUDE_CONTROL_SOURCE_SHIFT)) {setPitchAngleSetpoint(altitudeControl(getAltitudeInput(ALTITUDE_GS_SOURCE), getAltitude()));setAltitudeSetpoint(getAltitudeInput(ALTITUDE_GS_SOURCE));setThrottleSetpoint(throttleControl(getAltitudeInput(ALTITUDE_GS_SOURCE),getAltitude()));}
```
sergei1152: space out and add line breaks for sanity sake (I know it was like this before)

Autopilot/AttitudeManager/FixedWing.c
```diff
@@ -0,0 +1,248 @@
...
+
+    //Mixing!
+    outputMixing(outputSignal, &control_Roll, &control_Pitch, &control_Throttle, &control_Yaw);
+    
+    //Error Checking
+    checkLimits(outputSignal);
+    //Then Output
+    unsigned int cameraCounter = 0; //TODO: TEMPORARY, STORE TIME NOT COUNTER (OR BOTH) IMPLEMENT THIS A BETTER WAY
+    cameraPollingRuntime(getLatitude(), getLongitude(), getTime(), &cameraCounter, getRoll(), getPitch());
+    cameraGimbalStabilization(getRoll());
```
sergei1152: remove all this camera stuff, since we dont use any of it

sergei1152: you can remove the files relating to the camera stuff as well (in a seperate PR)

Autopilot/AttitudeManager/InputCapture.c
```diff
@@ -70,7 +69,7 @@ unsigned int* getICValues(unsigned long int sys_time)
      */
     if ((sys_time - ppm_last_capture_time) > PWM_ALIVE_THRESHOLD){
         int i = 0;
-        for (i = 0; i < PPM_CHANNELS; i++){
+        for (; i < PPM_CHANNELS; i++){
```
sergei1152: same thing for the for loop

Autopilot/AttitudeManager/InputCapture.h
```diff
@@ -19,10 +19,11 @@
...
 
 /**
- * If the PPM signal is inverted. Some receivers, such as the ezUHF have an inverted
- * PPM signal, so set this to 1. If using the OrangeRX, set this to 0
+ * If the PPM signal is inverted (inverted = active low). Some receivers, such 
+ * as the OrangeRX have an inverted PPM signal, so set this to 1. If using the 
+ * ezUHF, set this to 0
  */
-#define PPM_INVERTED 0
+#define PPM_INVERTED 1 // Might not be necessary? Single edge mode will catch either pulse
```
sergei1152: I dont like having unactionable comments like this in the code base. Delete the comment and create a github issue or an asana task to test whether this flag does anything by testing with the ezuhf. You dont have to do it but create it so that its tracked

Autopilot/AttitudeManager/Multirotor.c
```diff
@@ -0,0 +1,212 @@
+/*
+ * File:   Multirotor.c
+ * Author: Ian Frosst
+ *
+ * Created on February 13, 2017, 8:58 PM
+ */
```
sergei1152: Same comments as the fixed wing apply to this file

Autopilot/AttitudeManager/PWM.h
```diff
@@ -31,18 +31,32 @@
...
+#define UPPER_PWM 1250
+#define LOWER_PWM 625
 
 /**
- * Used in some of the calculations
+ * The middle of the PWM range of the RC controller. This is used as the initial
+ * offset in the output and input scaling calculations
  */
-#define HALF_PWM_RANGE (MAX_PWM - MIN_PWM)/2
+#define MIDDLE_PWM (int)(((UPPER_PWM - LOWER_PWM)/2) + LOWER_PWM)
```
sergei1152: Why were these defines moved to the header file? There not meant to be configurable, only calculated

Autopilot/AttitudeManager/PWM.h
```diff
@@ -84,6 +98,9 @@ int* getPWMArray(unsigned long int sys_time);
  */
 void setPWM(unsigned int channel, int pwm);
 
+// TODO: doc
```
sergei1152: document this

Autopilot/AttitudeManager/StateMachine.c
```diff
@@ -24,14 +23,10 @@ int downlinkP1Timer = 0;
 int downlinkP2Timer = 0;
 int imuTimer = 0;
 char AMUpdate = 0;
+char flightUpdate = 0;
```
sergei1152: should be static. Also rename to `updateFlight` and add comments describing what this variabe does

Autopilot/AttitudeManager/main.h
```diff
@@ -16,7 +16,7 @@
 
 //Turns debug mode on or off. Usually contains small snippets of code to output
 //data through UART or to provide small input adjustments
-#define DEBUG 0
+#define DEBUG 1
```
sergei1152: set this to 0

Autopilot/AttitudeManager/timer.c
```diff
@@ -36,10 +36,10 @@ void initTimer4(){
     T4CONbits.TGATE = 0; // Disable Gated Timer mode
     T4CONbits.TCKPS = 0b10; // Select 1:64 Prescaler
     TMR4 = 0x00; // Clear timer register
-    PR4 = 642; // Load the period value
-    IPC6bits.T4IP = 0x01; // Set Timer1 Interrupt Priority Level
-    IFS1bits.T4IF = 0; // Clear Timer1 Interrupt Flag
-    IEC1bits.T4IE = 1; // Enable Timer1 interrupt
+    PR4 = 625; // Load the period value
```
sergei1152: Theres a flag for this value. I beleive i named it T2_TIMER_TICKS_TO_MS however it applies to both timers. rename and adjust that define and set it to PR4, since what you're doing only updates Timer 4, but not Timer 2

Autopilot/GPS/Makefile
```diff
@@ -0,0 +1,113 @@
+#
```
sergei1152: get rid of this file

Autopilot/AttitudeManager/FixedWing.c
ianjfrosst: Brackets improve readability. Plus, this is all going away in the next PR I make (rewriting orientation controller).

Autopilot/GPS/Makefile
ianjfrosst: ...? You need makefiles.

Autopilot/AttitudeManager/timer.c
ianjfrosst: I felt that T2_TICKS_TO_MSEC should be used for Timer2 only, in the event that the timers are set with different prescale values (could improve output response times). I will try to make a macro for this value, but that will still cause problems, as there's no simple way to convert from the prescaler value to a multiplier.

Autopilot/GPS/Makefile
sergei1152: isnt this generated by the ide

Autopilot/AttitudeManager/AttitudeManager.c
ianjfrosst: I will try to move all the flap control code to FixedWing.h. Will depend on the PID rewrite I'm working on.

Autopilot/AttitudeManager/timer.c
sergei1152: just another T4_TICKS_TO_MSEC defien works then. Just add a comment taht states this value depends on the presclaer

Autopilot/AttitudeManager/FixedWing.c
sergei1152: not really in this case.

Autopilot/AttitudeManager/FixedWing.c
ianjfrosst: This is required. The header files are only included if their flag is set, but the source files are all compiled and linked regardless, as they are still in the project.

Autopilot/AttitudeManager/FixedWing.c
sergei1152: unless youre doing complicated dereferencing, its easy enough to see the star. The brackets just make it look ugly

Autopilot/AttitudeManager/FixedWing.c
ianjfrosst: Also going away with next PR.

Autopilot/AttitudeManager/FixedWing.c
sergei1152: ah ok

Autopilot/AttitudeManager/PWM.h
ianjfrosst: #defines should be in the headers. More efficient, as other files include PWM.h and use these values.

Autopilot/AttitudeManager/StateMachine.c
ianjfrosst: OK, but why `updateFlight` over `flightUpdate`?

sergei1152: i think having verbs first makes it sound more natural. like if(updateFlight) vs if(flightUpdate)

Autopilot/AttitudeManager/PWM.h
sergei1152: not necessarily. Defines that you want other files to use should be in the headers. MiddlePWM i can understand, but the scaling values shouldnt be included here

Autopilot/AttitudeManager/FixedWing.c
sergei1152: if the next PR is going to be ready soon, close this one and split it up into multiple PR's like i mentioned. Make sure that the PR's are logically seperated

Autopilot/AttitudeManager/FixedWing.c
chrishajduk84: Yeah I think that was my bad... But good catch.

Autopilot/AttitudeManager/FixedWing.c
chrishajduk84: Agreed

Autopilot/AttitudeManager/FixedWing.c
chrishajduk84: Empty functions don't get compiled anyway, yet show the potential for expanded functionality... 

But I see your point... This isn't gonna happen in the near future. 

So yeah remove it I guess... I would accept this PR   even if you had it.

Autopilot/AttitudeManager/PWM.h
ianjfrosst: Upper & lower PWM? Aren't those used for mapping inputs to sane values?

Autopilot/AttitudeManager/FixedWing.c
chrishajduk84: Agree with serge here.

Autopilot/AttitudeManager/PWM.h
ianjfrosst: Thought it would be more consistent.

Autopilot/AttitudeManager/FixedWing.c
ianjfrosst: Ooooh the parentheses. Thought you meant the braces from the if statement.

Autopilot/GPS/Makefile
ianjfrosst: No, I tried removing them with the rest of the stuff. IDE didn't like it.

Autopilot/AttitudeManager/PWM.h
chrishajduk84: If you expect to never use the defines outside a file, they should be in the c file indeed. There is a few that you can probably make private.

Autopilot/AttitudeManager/ProgramStatus.c
```diff
@@ -52,20 +52,18 @@ void setProgramStatus(int status){
...
+    } else if (status == ARMING) {
         debug("Arming: Motor Startup Procedure Started");
-    else if (status == MAIN_EXECUTION){
+    } else if (status == MAIN_EXECUTION) {
         debug("Motor Startup Procedure Complete");
         debug("Autopilot Running...");
-    }
-    else if (status == WAITING_TO_DIE_MODE){
+    } else if (status == KILL_MODE_WARNING) {
         debug("I am contemplating suicide.");
```
chrishajduk84: Whoever did this... Lol

Autopilot/AttitudeManager/net.h
```diff
@@ -15,16 +15,16 @@ extern "C" {
 #endif
 
 /** Time in miliseconds for how often a P0(high priority) packet gets sent down. Default=300 **/
-#define P0_SEND_FREQUENCY 300 
+#define P0_SEND_FREQUENCY 200
```
chrishajduk84: Since this isn't a proper operating operating system, this might cause problems. Be careful for future datalink problems (especially with uplink)

Autopilot/AttitudeManager/timer.h
```diff
@@ -12,12 +12,12 @@
...
  * Number of Timer2 ticks in a millisecond. To calculate this, take:
  * 1/(frequencyCPU/Timer2PreScaler)*TICKS_TO_MSEC should equal close to 0.001 or 1 ms
- * In this case, (1/(41Mhz/64))*642 == ~1ms
+ * In this case, (1/(40 MHz/64))*625 == 1 ms
  * 
  * This value is used to accurately define the timer2 period (used for output pwm
  * frequency
  */
-#define T2_TICKS_TO_MSEC 642
+#define T2_TICKS_TO_MSEC 625
```
chrishajduk84: This should be verified with an oscilloscope?

Autopilot/GPS/Makefile
chrishajduk84: It needs to be removed from the commit.

chrishajduk84: You can regenerate the makefile any time, by right clicking on the project going to the properties, and then hitting ok.

ianjfrosst:
I agree with you about the PR size. This one kind of snowballed, because the goal went from "refactor vehicles" to "make the quad fly", and doing that required fixing other things. I was impatient, and didn't want to make fixes again on a new branch, PR, wait for review, and pull in from master.

.gitignore
```diff
@@ -3,10 +3,8 @@ dist/
 *.d
 private/
 *.properties
-*.xml
 *.mk
 *.bash
-Makefile
```
ianjfrosst: Isn't the directory structure in the git repo?

Autopilot/GPS/Makefile
ianjfrosst: Did not know that. However, keeping the Makefiles does make it quicker to clone & set up the repo again.

Autopilot/AttitudeManager/PWM.h
sergei1152: Same reasoning as to why you would declare variables inside files static, you'd do the same with defines. It tells the programmer that the scope of that define only exists within this file, so he can rename it, modify it, etc without worrying adverse effects.

Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -847,11 +850,11 @@ void readDatalink(void){
                 break;
             case ARM_VEHICLE:
                 if (*(int*)(&cmd->data) == 1234)
-                    startArm();
+                    armVehicle(500);
```
ianjfrosst: armVehicle() is a function in either FixedWing.c or Multicopter.c. Both do different things.

Autopilot/AttitudeManager/timer.h
sergei1152: Is there a pin on the chip we can use for thtat? Whats the best way to physically test this

ianjfrosst: Should be verified, yes. However, by my math, I would be impossible for the previous value to be correct.

Autopilot/AttitudeManager/net.h
ianjfrosst: Have been keeping an eye out. May change back.

.gitignore
chrishajduk84: It is... The relative path is stored in git structure, but Makefiles use absolute path. This is true for make on any system. 

You can try it by copying and pasting it into a new folder, the old makefile won't work and will provide an error.

Autopilot/AttitudeManager/PWM.h
ianjfrosst: I only recently remembered about the static variables thing (after the PR). Have fixed those. Defines will be fixed soon.

.gitignore
chrishajduk84: And by copy and pasting, I mean if you move the project folder into the D:/ for example, it won't work.

Autopilot/AttitudeManager/AttitudeManager.c
chrishajduk84: Oh I see. My bad. I got confused.

Autopilot/AttitudeManager/timer.h
chrishajduk84: You'd have to output MiddlePWM  or something and see what you expect (probably 1ms pulse) 

The value seems close to what it was, so I don't expect too much of a problem.

sergei1152: im also not sure if the clock is exactly at 40Mhz. Theres the tolerance of the clock chips themselves, which we cant do anything about. However we actually use an 8Mhz clock that we postscale to 40Mhz. I havent done the post scaler calculations myself, but maybe theres a reason why they were set to 642? @chrishajduk84

ianjfrosst: I don't think he knows either.

chrishajduk84: It was based off of empirical data and then back calculated I think. 

But I'm not really sure

PR #98, opened 2017-03-05
Created new Logger module that uses new UART driver

sergei1152:
**Changes**
- Moved out ByteQueue and UART modules to the common folder as they are used in PathManager
- Refactored debug.h and debug.c and created Logger module to use the new asynchronous UART module
- ByteQueue now has a cap limit to how large it can expand
- ByteQueue now shrinks in size by 1/2 if its 1/4 full
- Can selectively turn on either TX or RX or both for a particular UART interface
- Modified some ByteQueue unit tests. Added unit tests for new logger module
- Uart interface init function takes in buffer size as a parameter, rather than using defines
- Logger now maintains a state of whether it was initialized. If not, will not output anything through UART. Therefore we dont need the `#if DEBUG` everywhere, we can just call the debug functions

**Note** Disabled UART2 interrupts that are defined in the new UART module as they will conflict with the network stuff (for now)


Autopilot/Common/Interfaces/UART.c
```diff
@@ -123,77 +128,95 @@ void initUART(unsigned char interface, unsigned long int baudrate)
...
+        }
 
-        initBQueue(&uart2_rx_queue, INITIAL_UART2_RX_BUFFER_SIZE);
-        initBQueue(&uart2_tx_queue, INITIAL_UART2_TX_BUFFER_SIZE);
+        U2MODEbits.UARTEN = 1; // And turn the peripheral on
+        U2STAbits.UTXEN = 1; //enable transmit operations (must be called after uart enable)
     }
 }
 
 void quoueTXData(unsigned char interface, unsigned char* data, unsigned int data_length)
```
chrishajduk84: quoue?

Autopilot/Common/Interfaces/UART.c
```diff
@@ -123,77 +128,95 @@ void initUART(unsigned char interface, unsigned long int baudrate)
...
-#endif
-
-#if INTERFACE_UART1_ENABLED
+//void __attribute__((__interrupt__, no_auto_psv)) _U2RXInterrupt(void)
+//{
+//    //while the rx buffer has characters available to read
+//    while (U2STAbits.URXDA) {
+//        pushBQueue(&uart2_rx_queue, U2RXREG);
+//    }
+//}
```
chrishajduk84: Why the comments?

Autopilot/Common/Interfaces/UART.c
sergei1152: omg no i spelled it incorrectly for so many things fukkkkkkkkkkkk

Autopilot/Common/Interfaces/UART.c
sergei1152: because it'll interfere with the uart2 interrupts defined in net_inbound and net_outbound

Autopilot/Common/Interfaces/UART.c
sergei1152: oh thank god its just that one instance

Autopilot/Common/Interfaces/UART.c
chrishajduk84: So why not remove it completely?

sergei1152: because this is valid code and will be activated with the network refactors

PR #100, opened 2017-03-10
Long live makefiles

ianjfrosst:
Somebody convinced me to remove the Makefiles. This has made a lot of people very angry and been widely regarded as a bad move.


chrishajduk84:
I don't like build files in git, but if you are absolutely sure you need it...

ianjfrosst:
But why not? A Makefile is platform-agnostic and does *not* include absolute paths. The .mk files do that, but those are automatically generated and ignored from this project. The only thing removing the Makefile does is make it harder to build the project.

sergei1152:
💯

PR #101, opened 2017-03-12
New Xbee Driver Implementation

sergei1152:
Before you freak out about the PR size, half of it is unit tests.

**Changes**
- New driver for Xbee that uses new uart interface
- Added git submodule for mavlink (however not used yet)
- Added info level messages to logger

**Driver Features:**
- Send AT commands to the Xbee to query its RSSI, Recieved Error count, and Transmission Error count
- Receive AT commands responses from the XBEE, which includes its xbee destination address, so we dont have to hardcode it in the code
- Ability to easily add transparent serial driver, so we can add support for 3DR radio modules, or use computer instead of datalink when testing

**Note** that this driver is not used anywhere yet, so putting it in the code base should not cause any issues. Next PR will implement this in the code (and be tested on the hardware)


Autopilot/AttitudeManager/Drivers/RadioXbee.c
```diff
@@ -0,0 +1,494 @@
...
+        }
+
+        uint16_t i;
+        for (i = 0; i < data_length; i++) {
+            copied[i] = data[i + 12];
+        }
+
+        *length = data_length;
+        return copied;
+        break;
```
sergei1152: good catch

sergei1152:
The point of this PR is less so to increase  speed, as the previous implementation was also asynchronous, but to increase reliability of datalink communications. As long as the Xbee communication is working fine, the code doesnt allow for missed uplinks due do a buffer overflow issue which we experienced before. It's also unit tested, so once I test it on the hardware, I'd feel alot more confident about flying with this than our current system.

PR #103, opened 2017-03-17
Redone SPI & DMA implementations

ianjfrosst:
Changes:
- Rewrote SPI implementations across both chips to use a common driver.
- Did the same with interchip DMA.

Next: 
- Add a generic IMU driver, with implementations for VN-100 and MPU-9250, possibly MPU-6050


Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -526,17 +529,17 @@ void imuCommunication(){
     imu_YawAngle = imuData[YAW];
     imu_PitchAngle = imuData[PITCH];
     imu_RollAngle = imuData[ROLL];
-#if DEBUG
+#if DEBUG && 0
```
chrishajduk84: Why is there a 0 here? This will never evaluate to true.

Autopilot/AttitudeManager/VN_lib.h
```diff
@@ -47,7 +47,7 @@
 #define VN_BYTE4(word)    ((unsigned char)((word) & 0x000000FF))
 
 /* Bit mask to get the nth byte out of a 32-bit word where n=0 is most significant and n=3 is least significant */
-#define VN_BYTE(word, n)   ((unsigned char)((word & (0x000000FF << (n*8))) >> (n*8)))
+#define VN_BYTE(word, n)   ((unsigned char)((word & (0x000000FF << (n*8))) >> (n*8))) // doesn't work?
```
chrishajduk84: Should work unless its bigger than a byte, then this function wouldn't work.

What's going on here?

chrishajduk84: Remove it if you know it doesn't work.

Autopilot/AttitudeManager/AttitudeManager.c
ianjfrosst: The code below it used to be commented out. I used it to debug the IMU, and this was easier.

Autopilot/AttitudeManager/VN_lib.h
ianjfrosst: It should work, yes. I've tested it on my system. Couldn't get it to work here. Will remove.

PR #106, opened 2017-03-22
PWM Disconnect detection on relay chip

SafwanQ:
-Initiated new Timer
-Recorded time values using TMR1
-Recorded at every edge
-Compared time since last edge to max value
-Can change max value (Currently set at 100)


Autopilot/Relay/AutoRelay_v0_01.c
```diff
@@ -77,6 +86,8 @@ unsigned int timePeriod1 = 0;         // Used in interrupt
 unsigned int timePeriod2 = 0;
 unsigned int t[3] = {0,0,0};                // Used in interrupt
 unsigned int i = 0;
+static unsigned long int lastCaptureTime; //Used in tracking time between edges
+unsigned long int thisCaptureTime;        //Used in tracking time between edges
```
ianjfrosst: This doesn't need to be global

Autopilot/Relay/AutoRelay_v0_01.c
```diff
@@ -139,6 +152,18 @@ void Init(void)
     IFS0bits.IC1IF = 0;         // Clear IC7 Interrupt Status Flag
     IEC0bits.IC1IE = 1;         // Enable IC7 interrupt
 
+    T1CONbits.TON = 0;          // Disable Timer
+    T1CONbits.TCS = 0;          // Select internal instruction cycle clock
+    T1CONbits.TGATE = 0;        // Disable Gated Timer mode
+    T1CONbits.TCKPS = 0b10;     // Select 1:64 Prescaler
+    
+    PR1 = 0xFFFF;               //Load the Period Value
+    TMR1 = 0x00;                //Clear Timer 1 Interrupt flag
```
ianjfrosst: This is the Timer1 register (where it holds the counter), not the interrupt flag.

Autopilot/Relay/AutoRelay_v0_01.c
```diff
@@ -254,9 +279,11 @@ int main (void)
...
 	while(1)
 	{
-
-                //The window on the controller is 43%-50%
-		if ((ch8_position > 400) && (ch8_position < 430)) //Controller Setting is: +47%, -33% //(ch8_position > 450)
+        thisCaptureTime = TMR1;
+        unsigned int timeDiff = (thisCaptureTime - lastCaptureTime)/T1_TICKS_TO_MSEC;
+                              
+               //The window on the controller is 43%-50%                                           //Controller Setting is: +47%, -33% //(ch8_position > 450)
+		if ((ch8_position > 400) && (ch8_position < 430) && (timeDiff <= PWM_ALIVE_THRESHOLD))     //Also that it hasn't been too long since last edge which means disconnect
```
ianjfrosst: What is with the indentation here?

Autopilot/Relay/AutoRelay_v0_01.c
```diff
@@ -139,6 +152,18 @@ void Init(void)
...
 
+    T1CONbits.TON = 0;          // Disable Timer
+    T1CONbits.TCS = 0;          // Select internal instruction cycle clock
+    T1CONbits.TGATE = 0;        // Disable Gated Timer mode
+    T1CONbits.TCKPS = 0b10;     // Select 1:64 Prescaler
+    
+    PR1 = 0xFFFF;               //Load the Period Value
+    TMR1 = 0x00;                //Clear Timer 1 Interrupt flag
+    
+    IFS0bits.T1IF = 0;
```
ianjfrosst: This one's the interrupt flag.

Autopilot/Relay/AutoRelay_v0_01.c
```diff
@@ -254,9 +279,11 @@ int main (void)
     Init();
 	while(1)
 	{
-
-                //The window on the controller is 43%-50%
-		if ((ch8_position > 400) && (ch8_position < 430)) //Controller Setting is: +47%, -33% //(ch8_position > 450)
+        thisCaptureTime = TMR1;
+        unsigned int timeDiff = (thisCaptureTime - lastCaptureTime)/T1_TICKS_TO_MSEC;
```
sergei1152: Because you're not taking into account for timer overflow, this code will produce unexpected results as you're doing a substraction between unsigned ints. Since you're not reseting the timer anywhere, eventually thisCaptureTime  may be smaller than lastCaptureTime, resulting in an unsigned integer overflow. You'll get a huge time diff and the relay will turn to manual when it shouldnt

Autopilot/Relay/AutoRelay_v0_01.c
```diff
@@ -77,6 +86,8 @@ unsigned int timePeriod1 = 0;         // Used in interrupt
 unsigned int timePeriod2 = 0;
 unsigned int t[3] = {0,0,0};                // Used in interrupt
 unsigned int i = 0;
+static unsigned long int lastCaptureTime; //Used in tracking time between edges
```
sergei1152: this variable needs to be declared as volatile, since it is set in an interrupt service routine and read outside int he main code

ianjfrosst:
Another method of doing this would be to just use a timer interrupt. Use a large pre-scale value (1:256), and have the timer period set to the timeout value. A flag is set whenever you get an timer interrupt, and is reset (along with the timer) whenever you get an IC interrupt. Then you could just check the flag in the `while()` loop.

sergei1152:
That'll work, but as you mentioned, you can only go up to 500ms.

ianjfrosst:
But when would you want to wait more than 500 ms before shutting off the relays?

sergei1152:
Yeah I suppose. Up to Safwan, I'm fine with both methods considering the simplicity of the relay

SafwanQ:
Setup global variable that stores time since start. But the clock speed that timer uses is tbd, so TICKS_TO_MSEC might be wrong

SafwanQ:
Should be done now, please review, thanks

sergei1152:
The code looks good but it cant be merged in until its tested

Autopilot/Relay/AutoRelay_v0_01.c
```diff
@@ -69,11 +70,11 @@
 /**
  * Number of Timer1 ticks in a millisecond. To calculate this, take:
  * 1/(frequencyCPU/Timer2PreScaler)*TICKS_TO_MSEC should equal close to 0.001 or 1 ms
- * In this case, (8 MHz/64)*1ms = 125
+ * In this case, (2 MHz/6)/1000 = 250
```
sergei1152: prescaler is 8 not 6

PR #107, opened 2017-03-23
Rewrite of orientation controller

ianjfrosst:
This is not final. There a still a few bugs to work out, and features to implement. Branch was getting too big, so I'm submitting this for review & feedback.

Changes: 
- Added a generic PID module
- Rewrote orientation controller entirely, using the new PID module
- Refactoring flaps to be more contained in the FixedWing module
- Added a way to dynamically limit rate & angle setpoints

To-do:
- Fix problems with quadcopter control
- Implement an airspeed PID (requires calibrated airspeed sensor)
- Add flaps to autopilot control
- Convert PIDs to fixed-point (for speed)
- Add new gain control & reading to datalink chain (net, data relay, ground station)


Autopilot/AttitudeManager/OrientationControl.c
```diff
@@ -1,146 +1,140 @@
...
+    if (delta_msec > 0) { // only compute time-sensitive control if time has elapsed
+        float dTime = delta_msec / 1000.0f; // elapsed time in seconds
+        
+        if (fabsf(pid->Ki) > 0) { // Integral control
+            pid->integral += (pid->Ki * error) * dTime;
+            
+            if (pid->integral < -pid->imax) { // ensure integral stays manageable
+                pid->integral = -pid->imax;
+            } else if (pid->integral > pid->imax) {
+                pid->integral = pid->imax;
```
sergei1152: i feel like this is something that should reset after a while, or be a rolling sum. otheriwse it'll always get stuck at the max integral value, no matter what happens throughout the flight

Autopilot/AttitudeManager/OrientationControl.c
```diff
@@ -1,146 +1,140 @@
...
+            }
+            output += pid->integral;
+        }
 
-    float error = setpoint - output;
-    if (integralFreeze == 0){
-        sum_gain[type] += error;
+        if (fabsf(pid->Kd) > 0) { // Derivative control
+            float derivative = (error - pid->lastErr) / dTime;
```
sergei1152: any particular reason you're using seconds for this? Is it just to keep the gain values at reasonable ranges?

Autopilot/AttitudeManager/OrientationControl.c
```diff
@@ -1,146 +1,140 @@
...
-    if (type == GAIN_KD){
-        return kd_gain[YPRH];
-    }
-    else if (type == GAIN_KP){
-        return kp_gain[YPRH];
-    }
-    else if (type == GAIN_KI){
-        return ki_gain[YPRH];
+
+float getGain(PID_channel channel, PID_gain type){
```
sergei1152: not sure if i like the inconsistent accessor interface. Some of the functions here take a pointer to a PID_val and this one takes in a channel?

Autopilot/AttitudeManager/Multirotor.c
```diff
@@ -42,10 +42,10 @@ void armVehicle(int delayTime){
     asm("CLRWDT");
     Delay(delayTime);
```
sergei1152: we should probably put the clear watchdog assembly isntruction within the delay function. Up to you if you want to do it in this PR, it'll clean up this code however

Autopilot/AttitudeManager/Multirotor.c
```diff
@@ -69,138 +69,101 @@ void dearmVehicle(){
...
-    if (getControlPermission(ALTITUDE_CONTROL,ALTITUDE_CONTROL_ON,ALTITUDE_CONTROL_SHIFT) && getControlPermission(ALTITUDE_CONTROL_SOURCE,ALTITUDE_GS_SOURCE,ALTITUDE_CONTROL_SOURCE_SHIFT)) {
-        setPitchAngleSetpoint(altitudeControl(getAltitudeInput(ALTITUDE_GS_SOURCE), getAltitude()));
-        setAltitudeSetpoint(getAltitudeInput(ALTITUDE_GS_SOURCE));
-        setThrottleSetpoint(throttleControl(getAltitudeInput(ALTITUDE_GS_SOURCE),getAltitude()));
+   
+    uint8_t rollControlType = getControlValue(ROLL_CONTROL_TYPE);
+    uint8_t rollControlSource = getControlValue(ROLL_CONTROL_SOURCE);
+    if (rollControlType == ANGLE_CONTROL) {
+        setRollAngleSetpoint(getRollAngleInput(rollControlSource));
+        setRollRateSetpoint(PIDcontrol(getPID(ROLL_ANGLE), getRollAngleSetpoint() - getRoll(), MAX_ROLL_RATE / MAX_ROLL_ANGLE));
```
sergei1152: can you explain (in a comment as well) why you're using the rate/angle for the scale factor of the pid?

Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -604,26 +508,21 @@ int coordinatedTurn(float pitchRate, int rollAngle){
...
-    return yawPID;
+// Type is both bit shift value and index of bit mask array
+uint8_t getControlValue(ctrl_type type) {
+    return (controlLevel & ctrl_mask[type]) >> type;
 }
 
-char getControlPermission(unsigned int controlMask, unsigned int expectedValue, char bitshift){
-    int maskResult = (controlMask & controlLevel);
-    return (maskResult >> bitshift) == expectedValue;
+void constrain(int16_t* input, int16_t min, int16_t max) {
```
sergei1152: this is a pretty general function. should maybe go into a seperate file (utilities.c?)

Autopilot/AttitudeManager/main.h
```diff
@@ -18,7 +18,7 @@
  * Whether to enable the logger. Disabling it will cause all the functions defined
  * in the module to do nothing
  */
-#define DEBUG 0
+#define DEBUG 1
```
ianjfrosst: Will do when it's ready to merge

Autopilot/AttitudeManager/StateMachine.c
```diff
@@ -31,15 +27,14 @@ static char flightUpdate = 0;
 void StateMachine(char entryLocation){
     //Timers
     dTime = (int)(getTime() - stateMachineTimer);
-    stateMachineTimer = getTime();
+    stateMachineTimer += dTime;
```
ianjfrosst: ...it's used to get the delta time, on the line right above this one.

Autopilot/AttitudeManager/OrientationControl.h
```diff
@@ -1,233 +1,107 @@
...
+#define	ORIENTATIONCONTROL_H
+
+#define PID_CHANNELS 8 // to start
+
+#define PID_RESET_TIME 1000 // timeout to reset I and D terms (ms)
+
+#define MAX_ROLL_ANGLE 35.f // degrees
+#define MAX_PITCH_ANGLE 35.f
+
+#define MAX_ROLL_RATE 180.f // degrees/second
```
ianjfrosst: Not really. There were max roll & pitch angles before, these are based off those + other flight controllers.

Autopilot/AttitudeManager/OrientationControl.h
```diff
@@ -1,233 +1,107 @@
...
+
+/**
+ * Sets a gain value for a PID channel
+ * @param channel Channel to set gain for
+ * @param type Type of gain (KP, KI or KD)
+ * @param value Value to set the gain to
+ */
+void setGain(PID_channel channel, PID_gain type, float value);
+
+uint8_t areGainsUpdated();
```
ianjfrosst: It's used as a trigger for one of the datalink packets.

Autopilot/AttitudeManager/OrientationControl.c
ianjfrosst: The ones that take a pointer are for generic PID control. The ones that are specific to the orientation controller take one of the channels. Need to make this more clear. Might separate into a PID.c file.

Autopilot/AttitudeManager/OrientationControl.c
ianjfrosst: The integral control does reset of it's not updated for more than a second (configurable). Getting stuck at the max integral value means you'd need a larger integral gain.

Autopilot/AttitudeManager/OrientationControl.c
```diff
@@ -1,146 +1,140 @@
...
-    //Derivative Calculations ---Not necessarily needed for altitude
-    int dValue = error - lastError[ALTITUDE];
-    lastError[ALTITUDE] = error;
+static uint8_t gainsUpdated = 0; // updated gain flag
 
 
-    int controlSignal = (int)(ALTITUDE_PITCH_SCALE_FACTOR * ((dValue * kd_gain[ALTITUDE]) + (error * kp_gain[ALTITUDE]) + (sum_gain[ALTITUDE] * ki_gain[ALTITUDE])));
-    return controlSignal;
+// Initial PID gains. These are only used to keep sane values on startup.
+const static float init_kp[PID_CHANNELS] = {1, 1, 1, 1, 1, 1, 1, 1};
```
ianjfrosst: Good idea. Different vehicles will have different PID response. Unfortunately, as the structure of the controller is different, those gains aren't as valid anymore. Will see if I can convert them to this structure.

Autopilot/AttitudeManager/StateMachine.c
sergei1152: ah lol

Autopilot/AttitudeManager/OrientationControl.h
sergei1152: :+1:

Autopilot/AttitudeManager/OrientationControl.h
sergei1152: ah thats right

Autopilot/AttitudeManager/OrientationControl.c
sergei1152: yeah i'd agree with a seperate file. Seperating stateless modules and stateful ones is always good, otherwise you get a file like attitude manager

Autopilot/AttitudeManager/OrientationControl.c
sergei1152: ok yeah. i forgot that the error can also be negative and decrease the value

Autopilot/AttitudeManager/OrientationControl.c
ianjfrosst: Pretty much, yeah. With milliseconds, You'd need really big KD values and really small KI ones.

Autopilot/AttitudeManager/Multirotor.c
ianjfrosst: I'm on the fence about the scale factor. Instead of using that, you could just multiply the output when you call the function. What it does is make it simpler to scale from an input to an output, so you wouldn't have to scale all the gains.

Autopilot/AttitudeManager/Multirotor.c
ianjfrosst: Out of scope.

Autopilot/AttitudeManager/OrientationControl.c
```diff
@@ -1,146 +1,140 @@
...
+ * The larger this is, the more twitchy D control is. 
+ */
+#define FILTER 0.4f
+
+/* 
+ * Floating-point PID loops for attitude control. 
+ * Inspired by ArduPilot control code.
+ * TODO: convert to fixed-point (integers) for increased speed, decreased size.
+ * (could probably do with tenth- or hundredth-degree integers)
+ * Floating-point math on the dsPIC33 is done via emulation, and is very slow.
```
ianjfrosst: Was looking at options. There's -relaxed-math, with does exactly that, and there's -fast-math, that does even faster floating point math while maintaining precision, but at the cost of size.

Autopilot/AttitudeManager/Multirotor.c
sergei1152: I'm not sure why we need a scale factor? Isnt it the same operation as multiplying all the gains by the given scale factor? In which case why complicate the control system with it

Autopilot/AttitudeManager/OrientationControl.h
```diff
@@ -81,27 +49,27 @@ void orientationInit();
...
+PIDVal* getPID(controlChannel channel);
 
 /**
  * Gets a gain value from a PID channel
  * @param channel Channel to get gain from
  * @param type Type of gain (KP, KI or KD)
  * @return Requested gain value
  */
-float getGain(PID_channel channel, PID_gain type);
+float getGain(controlChannel channel, gainType type);
```
sergei1152: isnt the naming scheme capital camel case? ControlChannel, etc

Autopilot/AttitudeManager/Multirotor.c
chrishajduk84: @sergei1152 It scales the ENTIRE output signal rather than just the proportional or derivative gains.
You could probably get rid of it, but it makes it convenient to have reasonable numbers for tuning the system.

Autopilot/AttitudeManager/PID.c
```diff
@@ -0,0 +1,74 @@
...
+        delta_msec = 0;
+        pid->integral = 0;
+        pid->last_err = error;
+    }
+    
+    pid->last_time = now;
+
+    output += pid->kp * error; // Proportional control
+    
+    if (delta_msec > 0) { // only compute time-sensitive control if time has elapsed
```
chrishajduk84: Line 47 is also time-sensitive. If delta_msec is zero, it is likely nothing has changed in that time. As a result, the error may become zero when it should not.

Autopilot/AttitudeManager/PID.c
```diff
@@ -0,0 +1,74 @@
...
+                pid->integral = -pid->i_max;
+            } else if (pid->integral > pid->i_max) {
+                pid->integral = pid->i_max;
+            }
+            output += pid->integral;
+        }
+
+        if (fabsf(pid->kd) > 0) { // Derivative control
+            float derivative = (error - pid->last_err) / dTime;
+            derivative = derivative * FILTER + pid->last_der * (1-FILTER); // reduce jitter in derivative by averaging
```
chrishajduk84: Theoretically not a PID loop anymore...

Autopilot/AttitudeManager/PID.h
```diff
@@ -0,0 +1,57 @@
...
+ * @copyright Waterloo Aerial Robotics Group 2017 \n
+ *   https://raw.githubusercontent.com/UWARG/PICpilot/master/LICENCE 
+ */
+
+#ifndef PID_H
+#define	PID_H
+
+#include "main.h"
+
+#define PID_RESET_TIME 1000 // timeout to reset I and D terms (ms)
```
chrishajduk84: What?

Autopilot/AttitudeManager/net.h
```diff
@@ -15,13 +15,13 @@ extern "C" {
 #endif
 
 /** Time in miliseconds for how often a P0(high priority) packet gets sent down. Default=300 **/
-#define P0_SEND_FREQUENCY 200 
+#define P0_SEND_FREQUENCY 300
```
chrishajduk84: Was it giving you problems?

Autopilot/AttitudeManager/PID.c
ianjfrosst: kp is not time sensitive. It's the instantaneous error.

Autopilot/AttitudeManager/PID.c
ianjfrosst: What do you mean?

Autopilot/AttitudeManager/PID.h
ianjfrosst: If you go too long without updating the PID, it resets the integral and the last derivative. Prevents integral windup and derivative spikes when on the ground.

Autopilot/AttitudeManager/net.h
ianjfrosst: No, just realized it shouldn't have been changed. Didn't realize it was in this PR. Thought it was the last one. Will revert. Serge's PR changes this anyway.

Autopilot/AttitudeManager/Multirotor.c
sergei1152: ok sounds good then

ianjfrosst: @chrishajduk84 it technically is the same as scaling all the gains individually, but makes for a bit easier tuning. There are a few of these features in the PID code, just for simpler gain tuning.

Autopilot/AttitudeManager/AttitudeManager.c
ianjfrosst: Could move it to Common?

sergei1152: the common folder? yes, not the Common.c file tho

Autopilot/AttitudeManager/PID.c
ianjfrosst: @chrishajduk84 ?

Autopilot/AttitudeManager/PID.c
chrishajduk84: What I mean by time sensitive is that if the processing time (loop time) varies, your response does as well....

Details... I'm just trying to nitpick

Autopilot/AttitudeManager/PID.c
chrishajduk84: NVM i misread something.

Also here...why bother with fabsf()?

Negative gains are completely valid.

Autopilot/AttitudeManager/PID.h
chrishajduk84: Got it cool.

Autopilot/AttitudeManager/PID.c
ianjfrosst: Exactly. Check if the ABSOLUTE gain is greater than 0.

ianjfrosst: So if a gain exists that's not zero, it'll run the code.

chrishajduk84: I'm an idiot. Thank you for pointing that out.

PR #108, opened 2017-03-26
New Network manager implementation using xbee driver

sergei1152:
**Changes**
- `net.h` has been renamed to `Datalink.h`
- `net_outbound.c` , `net_common.c`, and `net_inbound.c` have been removed and replaced with `Datalink.c`
- removed `UART2.h` and `UART2.c`
- added a couple of helper methods in the `Logger` module for printing out integers and arrays without manually doing an sprintf everytime
- retrieving the xbee destination address dynamically now, instead of hardcoding it. This way, we can interchange pre-configured xbees without having to reflash the pic (will be required and useful for quad)
- fixed checksum stuff in the radio driver
- added 100ms delay on picpilot startup since when plugging in the picpilot, you get an inconsistnet supply of power that makes the coms between the xbee inconsistent
- saved about 2k of ram/data memory since we're no longer statically allocating for the staging queue stuff that was implemented before

**Note**
I've done a full system test with this and it works. Got a data rate of 5-7/s going, with no uplink transmissions being missed.

Alot of the unit tests I wrote for this are broken due to the checksum changes. Upcoming PR will fix those as well as add unit tests for datalink module. Next PR will also add the xbee radio packet stats (such as RSSI) in the down link


Autopilot/AttitudeManager/main.h
```diff
@@ -18,7 +18,7 @@
  * Whether to enable the logger. Disabling it will cause all the functions defined
  * in the module to do nothing
  */
-#define DEBUG 0
+#define DEBUG 1
```
ianjfrosst: DEBUG = 0 for master branch

Autopilot/AttitudeManager/nbproject/configurations.xml
```diff
@@ -274,6 +266,468 @@
...
+        <property key="warningmessagebreakoptions.W9803_SCL_FAILED_TO_READ_FILE"
+                  value="report"/>
+        <property key="warningmessagebreakoptions.W9804_SCL_UNRECOGNIZED_LABEL"
+                  value="report"/>
+        <property key="warningmessagebreakoptions.W9805_SCL_UNRECOGNIZED_VAR"
+                  value="report"/>
+        <property key="warningmessagebreakoptions.displaywarningmessagesoption"
+                  value=""/>
+        <property key="warningmessagebreakoptions.warningmessages" value="holdstate"/>
+      </Simulator>
```
ianjfrosst: Is this necessary?

Autopilot/Common/Utilities/Logger.c
```diff
@@ -55,6 +62,31 @@ void debug(char* message)
...
+        if (to_print == NULL){
+            return;
+        }
+        int i;
+        for (i = 0; i < length; i++) {
+            sprintf(&to_print[i*3],"%02x ", array[i]);
+        }
+        to_print[length*3] = 0;
+
+        writeMessage((char*)to_print, DEBUG_TAG_STRING, DEBUG_TAG_STRING_LENGTH);
```
ianjfrosst: to_print is already of type (char*), why the cast?

Autopilot/AttitudeManager/nbproject/configurations.xml
ianjfrosst: (the 400+ lines added)

ianjfrosst:
Looks good. Much cleaner than before.

CoatedMoose:
Using dynamic memory allocation was explicitly avoided when I initially wrote this. The problem with malloc and free is that tracking what memory is available is inevitably slower, or memory inefficient. Either you need complex data structures (more memory, lots of memory fragmentation), wasteful memory allocation algorithms (mark a block as used, never free it, or wait for entire block to be emptied), or complex algorithms on allocation/free/both. None of these are great options for one reason or another, and is entirely dependent on the systems implementation. 

As malloc and free are called more often, fragmenting WILL grow, resulting in less "actually free" memory, as well as malloc and free taking longer. If this extends beyond the time to fill the UART buffer (the UART interrupt fires), you will start losing and corrupting telemetry. The other concern is you start dropping PID cycles, which is bad for flight stability.

This is not too say the change is necessarily bad, but it is important to be aware of the potential side effects of relying of dynamic memory allocation.

Autopilot/AttitudeManager/Drivers/RadioXbee.c
```diff
@@ -139,24 +143,22 @@ void queueRadioStatusPacket()
 
 void clearRadioDownlinkQueue()
 {
-    XbeeApiFrame* packet = XbeeFrameQueue.head;
+    XbeeApiFrame* packet = popApiFrame();
     while (packet != NULL) {
-        XbeeFrameQueue.head = packet->next_frame;
-        free(packet->data);
-        free(packet);
```
sergei1152: oh good catch. I deleted that code blindly without looking at it

Autopilot/AttitudeManager/Drivers/RadioXbee.c
```diff
@@ -268,6 +276,9 @@ uint8_t* parseUplinkPacket(uint16_t* length)
 
     //the parsed length of the packet we're parsing. Used to tell when the packet is complete
     static uint16_t rx_packet_length;
+    
+    //current calculated checksum of the in-progress packet
+    static uint8_t checksum;
```
sergei1152: Because when this method gets called we might not have received the entire xbee frame. This method must be capable of handling of parsing partial xbee frames, and so we need to control when the checksum gets reset, and also keep its state throughout the receival of the packet

Autopilot/AttitudeManager/nbproject/configurations.xml
sergei1152: folder structure differences, so yes

Autopilot/Common/Utilities/Logger.c
sergei1152: good point

sergei1152:
Yes there are advantages and disadvantages to both. I opted for dynamic allocation mainly to increase reliability without using more memory. I personally don't like allocating large segments of memory that may or may not be used, and would rather make that memory available for the heap. I'm aware of the speed hit for using malloc and free but time will tell if its even an issue (there are far slower operations that happen on the picpilot atm)

Autopilot/AttitudeManager/nbproject/configurations.xml
sergei1152: actually i dont know what the rest of this stuff is, ill delete it

Autopilot/AttitudeManager/Drivers/RadioXbee.c
ianjfrosst: Ah. Ok.

Autopilot/AttitudeManager/main.c
```diff
@@ -48,6 +48,11 @@ int main(int argc, char** argv) {
     AD2PCFGLbits.PCFG4 = 1;
     AD2PCFGLbits.PCFG5 = 1;
     
+    //as we plug in the picpilot, there may be intermittent power from the initial contact of the
+    //power plug which the sensor drivers don't like. This delay is meant to stop communication of sensors 
+    //until we know we're getting constant power
+    Delay(100);
```
chrishajduk84: Good idea.

sergei1152:
Realized line 195 in the xbee driver is gonna be a problem, since it skips over every 2nd packet that it has to send. surprised i didnt see a memory leak..

ianjfrosst:
I'd suggest doing some sort of endurance test before merging, to check for memory leaks that haven't been caught. Send commands and receive telemetry for longer than we'd ever have PICpilot on for, and if it can do that without crashing, I'd call it good.

Autopilot/AttitudeManager/Drivers/RadioXbee.c
```diff
@@ -192,7 +192,6 @@ bool sendQueuedDownlinkPacket()
         
         queueTXData(XBEE_UART_INTERFACE, send_buffer, total_frame_length); //queue the data for tranmission over UART
 
-        XbeeFrameQueue.head = to_send->next_frame;
```
ianjfrosst: Why the change?

sergei1152: the pop command already does this

ianjfrosst: Right.

sergei1152:
Yes I was thinking of doing an endurance test just to make sure. I'll do one tonight with a data rate of 5-7/s and see if there are any issues

sergei1152:
So after doing the test for 30min +, there didnt seem to be any interruptions in the data rate. I noticed however that when i send multiple uplinks, the data rate drops to about 2/s. This is actually what also happens in master, and it looks to be the result of the state machine timings. I'll work on that next and optimize it so that it doesn't starve other critical timings.

chrishajduk84:
I'm like 100% sure it's not statemachine timings. 

The xbee can only do uplink or downlink individually, but requires a few milliseconds to switch between modes. It essentially requires a hardware level handshake to accomplish. It takes time to do and that is what you are seeing. Code changes won't fix that.

sergei1152:
I suspect thats part of it. However I'll still want to add some checks to make sure that, if we write a P0 and P1 and P2 packet, that they dont get sent right after each other, which we currently dont do, and might cause other tasks to starve (might).

PR #110, opened 2017-04-08
Fixed parsing uplinks & added xbee and uhf rssi statistics to downlink

sergei1152:
**Changes**
- Used my powers to fix the master branch tree to not include like 40 unessary commits after the revert
- Added UHF module for interpreting the rssi and link quality channel inputs, and sending the data down the downlink
- Sending down xbee rssi, transmission errors, and received error count down the downlink. Refreshing values everytime a heartbeat is received
- Not allocating the Telemetry block dynamically in the writeDatalink function. Instead doing it statically for performance reasons
- Fixed parsing of multi-byte uplink packets. The issue was caused due to having improper byte alignment. For example, the arm command contains 3 bytes, the cmd id, and 2 bytes that are the password. Dereferencing the second entry in the array will give you an opcode reset, since the chip doesnt like to dereference things that aren't word aligned (need to be a multiple of 2). Fixed by copying everything in the command data after the command id to a new, word aligned array so that casts would work
- Still need to do further testing and maybe make the code cleaner but should work for now

Data Relay PR: https://github.com/UWARG/data-relay-station/pull/37

### DIFF
--- a/Autopilot/AttitudeManager/AttitudeManager.c
+++ b/Autopilot/AttitudeManager/AttitudeManager.c
@@ -516,6 +516,12 @@ uint8_t getControlValue(CtrlType type) {
 
 void readDatalink(void){
     struct DatalinkCommand* cmd = popDatalinkCommand();
+    uint8_t cmd_data[88];
+    
+    //this step is necessary for correct parsing of data, as we can't cast
+    //unaligned data to multi-byte values.
+    memcpy(cmd_data, cmd->data, cmd->data_length);
+    
     //TODO: Add rudimentary input validation
     if ( cmd ) {
         if (lastCommandSentCode[lastCommandCounter]/100 == cmd->cmd){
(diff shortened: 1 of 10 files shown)


chrishajduk84:
"For example, the arm command contains 3 bytes, the cmd id, and 2 bytes that are the password. Dereferencing the second entry in the array will give you an opcode reset, since the chip doesnt like to dereference things that aren't word aligned (need to be a multiple of 2). Fixed by copying everything in the command data after the command id to a new, word aligned array so that casts would work"

_Is it better to do that rather than just byte aligning it from the get go?_

Autopilot/AttitudeManager/Network/Datalink.h
```diff
@@ -14,16 +14,16 @@
...
 /** Time in miliseconds for how often a P1(medium priority) packet gets sent down. Default=1000 **/
-#define P1_SEND_FREQUENCY 800
+#define P1_SEND_FREQUENCY 1000
 
 /** Time in miliseconds for how often a P2(low priority) packet gets sent down. Default=20000 **/
 #define P2_SEND_FREQUENCY 5000
 
 /** Time in miliseconds for how often to check for new messages from the uplink. Default=100 **/
-#define UPLINK_CHECK_FREQUENCY 75
+#define UPLINK_CHECK_FREQUENCY 500
```
chrishajduk84: All of these frequencies have increased....is there a reason why?

chrishajduk84: Rather...frequency isn't the right word, period is what I meant.

Autopilot/AttitudeManager/Peripherals/UHF.c
```diff
@@ -0,0 +1,36 @@
+/* 
+ * @file UHF.h
+ * @author Serj Babayan
+ * @created March 26, 2017, 9:00 PM
+ */
+#include "../InputCapture.h"
+#include "../PWM.h"
+#include "UHF.h"
+
+uint8_t getUHFRSSI(uint32_t sys_time){
```
chrishajduk84: This scaling is sort of a duplicate of the scaling function available in PWM.c

Autopilot/AttitudeManager/Network/Datalink.h
sergei1152: This is sorta like a preventative measure against spamming commands. ie. if you're doing it, I dont want the datalink downlink speed to decrease as the picpilot parses stuff, which it will since 75 is lower than everything else. I'm still playing around with these values to find what I think is best.

Autopilot/AttitudeManager/Peripherals/UHF.c
sergei1152: Yeah I agree I think this is redudanant. I'll change it

sergei1152:
@chrishajduk84 So the reason for the byte alignment issue is that now the first byte in the command data field of the command structure contains the command id, which it didnt before. After thinking about it, I realized that since we're already storing the command id, i can overwrite the array and shift everything to the left by one, which I'll do in a change to this PR later. This will avoid the buffer thing that I had to do.

Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -714,175 +723,178 @@ void readDatalink(void){
...
+       freeDatalinkCommand( cmd );
     }
 }
 
 bool writeDatalink(p_priority packet){
-    TelemetryBlock* statusData = createTelemetryBlock(packet);
-
-    //If Malloc fails, then quit...wait until there is memory available
-    if (!statusData){return 0;}
+    static TelemetryBlock statusData;
```
ianjfrosst: Probably avoiding re-allocating it every time we send telemetry.

Autopilot/AttitudeManager/Peripherals/UHF.c
sergei1152: Wasn't actually able to get it to work using the PWM module for whatever reason. Couldn't get values to lie within the 0 - 100 range, probably doing something stupid. Will figure it out later as its low priority.

sergei1152:
Updated the PR to do the memcpy in the data link module, which makes more sense.

PR #112, opened 2017-04-20
Pruned old commands, new gain commands, new downlink format

sergei1152:
**Changes:**
- Added interchip dma errors counts to downlink
- Purged old set gain commands and replaced with new ones
- Removed priority based packet types. Now they are related based on the information they hold (ie. position, status, gains, channels)
- Modified state machine such that now there is a minimum interval of time between when packets get sent, and a queue that determines what packet should be sent. Now that we have 1 timer, this avoids issues when we had a priority 1 timer being a multiple of a priority 2 timer, causing them to send immediately one after the other
- Gains type packet now only sends on request, or if the gains were updated
- Fixed parsing of floats for gains in readDatalink() function
- Renamed lots of variables to match our naming convention
- Added ability to display non-scaled pwm outputs in datalink
- Purged alot of status variables that were being sent down. Replaced with new ones (ie. gps status replaced with peripheral status, etc)..

Corresponding datalink PR:
https://github.com/UWARG/data-relay-station/pull/38

### DIFF
--- a/Autopilot/AttitudeManager/AttitudeManager.c
+++ b/Autopilot/AttitudeManager/AttitudeManager.c
@@ -455,6 +455,10 @@ uint8_t getControlValue(CtrlType type) {
 void readDatalink(void){
     struct DatalinkCommand* cmd = popDatalinkCommand();
    
+    debugInt("position", sizeof(struct packet_type_position_block));
+    debugInt("status", sizeof(struct packet_type_status_block));
+    debugInt("gains", sizeof(struct packet_type_gain_block));
+    debugInt("channels", sizeof(struct packet_type_channels_block));
     //TODO: Add rudimentary input validation
     if ( cmd ) {
         if (lastCommandSentCode[lastCommandCounter]/100 == cmd->cmd){
@@ -663,7 +667,7 @@ void readDatalink(void){
     }
 }
 
-bool writeDatalink(p_priority packet){
+bool writeDatalink(PacketType packet){
     static TelemetryBlock statusData;
 
     int* input;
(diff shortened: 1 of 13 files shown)


Autopilot/AttitudeManager/AttitudeManager.c
```diff
@@ -693,137 +644,148 @@ void readDatalink(void){
...
+                setGain(ROLL_ANGLE, KP, CMD_TO_FLOAT(cmd->data));
+                setGain(ROLL_ANGLE, KD, CMD_TO_FLOAT(cmd->data + 4));
+                setGain(ROLL_ANGLE, KI, CMD_TO_FLOAT(cmd->data + 8));
                 break;
-            case SET_GAINS:
-            {
-                char channel = cmd->data[0];
-                //We're making the rest of the data word aligned here, which is required for casting it to floats
-                memcpy(cmd->data, cmd->data + 1, cmd->data_length - 1);
-                setGains(channel, CMD_TO_FLOAT_ARRAY(cmd->data));
```
ianjfrosst: Why is there no longer a way to set an arbitrary gain?

Autopilot/AttitudeManager/Network/Datalink.h
```diff
@@ -50,67 +63,72 @@ typedef enum _p_priority {
...
+    struct packet_type_position_block position_block;
+    struct packet_type_status_block status_block;
+    struct packet_type_gain_block gain_block;
+    struct packet_type_channels_block channels_block;
 } PacketPayload;
 
-typedef struct TelemetryBlock {
-    uint8_t type;
+typedef struct {
+    uint16_t type; // 2 bytes to avoid weird padding issues
```
ianjfrosst: Can you say anything more about "weird padding issues"?

Autopilot/AttitudeManager/Network/Datalink.h
```diff
@@ -34,12 +29,30 @@
...
+ * be called to move the packet type to the front of the queue
+ */
+static const uint8_t DEFAULT_PACKET_ORDER[] = {
+    PACKET_TYPE_POSITION,
+    PACKET_TYPE_POSITION,
+    PACKET_TYPE_POSITION,
+    PACKET_TYPE_STATUS,
+    PACKET_TYPE_POSITION,
+    PACKET_TYPE_CHANNELS
+};
```
ianjfrosst: Why is this order the way it is?

sergei1152: So that we get status and channel updates about once every second. I put the position in the middle of those two so that we also get the most updated positional information most of the time,

Autopilot/AttitudeManager/AttitudeManager.c
sergei1152: Because sending individual kp, ki, kd gains is pretty inefficient, if you have to set all of the gains of the aircraft (thats like 30 commands you have to send, with this theres 10). I dont want to have multiple commands doing the same thing as well

Autopilot/AttitudeManager/Network/Datalink.h
sergei1152: Ok i'll update it. I put this in because the data relay parses the first 2 bytes as the packet type, which we happened to get away with since the compiler always padded the TelemetryData block to take up an extra byte, which worked out

Autopilot/AttitudeManager/AttitudeManager.c
ianjfrosst: `setGains` sets all the gains for a specific channel. It's very useful, especially when tuning new loops.

Autopilot/AttitudeManager/Network/Datalink.h
ianjfrosst: Ok. But why not "pos, pos, stat, pos, pos, chan"? Why not make it symmetrical?

Autopilot/AttitudeManager/Network/Datalink.h
ianjfrosst: Ah. So it's not really a padding issue, just a mistake that was overlooked. That's OK.

Autopilot/AttitudeManager/AttitudeManager.c
sergei1152: are you talking about me removing the function? Because you can still set arbitrary gains for all the channels on the picpilot

ianjfrosst: All channels? How?

sergei1152: the `set_roll_rate_gains`, etc... commands

ianjfrosst: That's not arbitrary gains for all channels. What if I wanted to set Ki for altitude control? or add a little bit of Kd to the ground speed controller?

ianjfrosst:
LGTM 👍

PR #115, opened 2017-04-24
Path manager

dmurds:
Problem: Path manager was crashing after a single waypoint was inserted into it.
Cause: followWaypoint, the path calculation method at runtime, assumes that there are 3 waypoints in the queue that haven't been processed at any given moment. 

Solution: Added null checking to see how many waypoints past the current one are defined. If there aren't 3 waypoints defined, we default to a straight line path to the next waypoint.


sergei1152:
Can this be merged?

sergei1152:
@dmurds

PR #116, opened 2017-04-26
Fixed GPS

sergei1152:
**Changes:**
- Added driver support for multiple GPS's in the path manager
  - Implemented WARG GPS driver that operates over SPI (doesnt work yet, still need to investigate).
  - Implemented Gps driver for the ublox 6, but also using it for the WARG GPS when its configured to output NMEA strings over UART
- Refactored the GPS code to operate in SPI slave mode. Also reorganized everything, though  kept the NMEA parsing logic the same
- Added ability for GPS to output straight NMEA stirngs from its gps module once everything is configured. This is currently used by the path manager

- Modified baud rate calculation so that it rounds instead of cutting off the entire decimal place (required for proper communication with GPS)

TLDR: This PR is super long as I tried alot of stuff with the GPS. I couldn't get it to receive SPI interrupts (could be a hardware issue, need to further investigate). To get it to work I set it up to output NMEA strings directly out of its uart port, which feeds into the path manager. Used the existing NMEA parsing code in the path manager and re-implemented it as a driver for the ublox GPS, which was repurposed to work with our gps (was as simple as a baudrate change).

Out of my tests, I've noticed i got checksum errors from the NMEA strings a few times every 20-40 messages. It should be OK but further work is still required for greater reliability.


ianjfrosst:
Goddamn

Autopilot/Path Manager/Drivers/WARG_GPS.c
```diff
@@ -0,0 +1,148 @@
...
+
+static bool data_available = false;
+
+static bool connected = false;
+
+static void initDMA2(void);
+static void initDMA3(void);
+
+static volatile uint8_t dma2_space[sizeof(GPSData) + 2] __attribute__((space(dma))); //one byte for the checksum, and another for the synchronization byte
+static volatile uint8_t dma3_space[sizeof(GPSData) + 2] __attribute__((space(dma)));
```
ianjfrosst: Why channels 2 & 3?

Autopilot/Path Manager/PathManager.c
```diff
@@ -124,14 +118,20 @@ void pathManagerInit(void) {
...
 //    sprintf(&str,"%f",pmData.time);
 //    UART1_SendString(&str);
 #endif
-    copyGPSData();
+    requestGPSInfo();
 
+    if (isNewGPSDataAvailable()){
+        copyGPSData();
+    }
+
```
ianjfrosst: This will cause interchip to not update until there's new GPS data. This check should be moved into the `copyGPSData` function, for just the section using GPS data. `copyGPSData` might need to be renamed, too.

Autopilot/Path Manager/PathManager.c
```diff
@@ -640,18 +640,15 @@ unsigned int insertPathNode(PathData* node, unsigned int previousID, unsigned in
...
-    }
+    interchip_send_buffer.pm_data.time = gps_data.utc_time;
+    interchip_send_buffer.pm_data.longitude = gps_data.longitude;
+    interchip_send_buffer.pm_data.latitude = gps_data.latitude;
+    interchip_send_buffer.pm_data.heading = gps_data.heading;
+    interchip_send_buffer.pm_data.speed = gps_data.ground_speed;
+    interchip_send_buffer.pm_data.satellites = (char)gps_data.num_satellites;
+    interchip_send_buffer.pm_data.positionFix = (char)gps_data.fix_status;
+    
+    checkForFirstGPSLock();
```
ianjfrosst: As mentioned above, everything below here doesn't involve the GPS, and won't get updated if there's no new GPS data (e.g. the GPS module isn't connected).

Autopilot/GPS/MTK33x9.c
```diff
@@ -0,0 +1,385 @@
...
+            debug("Failed checksum when parsing a GPGGA (positional) packet!");
+            debug(gga_buffer);
+        }
+    }
+
+    if (new_vtg_data){
+        new_vtg_data = false;
+        if (isNMEAChecksumValid(vtg_buffer)){
+            data_available = false;
+            parseVTG(gga_buffer);
```
ianjfrosst: Shouldn't this use the VTG buffer?

Autopilot/GPS/UART.c
```diff
@@ -0,0 +1,144 @@
...
+#include <xc.h>
+#include "../Common/Utilities/ByteQueue.h"
+#include <stdint.h>
+
+/**
+ * The chip requires the baud rate register to be set to a value that corresponds
+ * to the processor frequency. The formula can be found on the data sheet. This
+ * takes an ordinary baud rate value in and returns that
+ */
+#define BRGVAL(bd) (((16000000/bd)/16) - 1)
```
ianjfrosst: Why use a magic number here if CLOCK_FREQUENCY is defined?

Autopilot/GPS/MTK33x9.c
```diff
@@ -0,0 +1,385 @@
...
+
+    i = 1;
+    long int multiplier = 10;
+    int decimalPoint = 0;
+    
+    gps_data.heading = 0;
+    float tHeading = 0;
+    for (i = 1; i < 6; i++) //this code first generates an 5 digit decimal number
+    {
+        if (rawHeading[i] == 0x10)//check for decimal point
```
sergei1152: I just copied that logic off of the existing code

Autopilot/GPS/MTK33x9.c
sergei1152: yes good point

Autopilot/GPS/LED.c
```diff
@@ -0,0 +1,24 @@
...
+ * @file LED.c
+ * @author Serj Babayan
+ * @created April 21, 2017, 12:43 AM
+ *
+ * @copyright Waterloo Aerial Robotics Group 2017 \n
+ *   https://raw.githubusercontent.com/UWARG/PICpilot/master/LICENCE
+ */
+
+#include <xc.h>
+#include <stdbool.h>
```
sergei1152: right..

Autopilot/GPS/MTK33x9.c
```diff
@@ -0,0 +1,385 @@
...
+        }
+    }
+    decimalPoint = decimalPoint - 2;
+    multiplier = 100000;
+    while (decimalPoint > 0) //then divdes it according to the placement of the decimal
+    {
+        multiplier = multiplier / 10;
+        decimalPoint--;
+    }
+    gps_data.ground_speed = gps_data.ground_speed / multiplier;
```
sergei1152: This is what we did in the old code. I dont think its worth changing at this moment (path manager expects m/s)

Autopilot/Path Manager/Drivers/WARG_GPS.c
sergei1152: 0 and 1 are taken up by interchip

Autopilot/Path Manager/PathManager.c
sergei1152: good pont

Autopilot/Path Manager/Drivers/WARG_GPS.c
ianjfrosst: Right. Thought this was on the GPS chip.

Autopilot/GPS/MTK33x9.c
ianjfrosst: I think the GPS gives speed in km/h. Will investigate. Not important to this PR.

Autopilot/GPS/MTK33x9.c
ianjfrosst: Ok.

chrishajduk84:
Probably not related, but there is errdata on the microchip website about a mistake in the hardware for SPI:
SPI - TBF bit 5. - Writing to the SPIBUF register as soon as the TBF bit is
cleared will cause the SPI module to ignore the written data.

Autopilot/GPS/MTK33x9.c
sergei1152: Yes the GPS module itself does, we do the conversion (i guess to avoid calculations?)

Autopilot/Path Manager/Drivers/WARG_GPS.c
sergei1152: Yeah GPS doesnt support DMA

sergei1152:
@chrishajduk84 I took a look at the errata for the pic24, and their SPI issues seems to be related to enchanced buffer mode, which I'm not using. The dspic33 one seems to not be an issue with DMA, which is what I'm using. I"ll do futher testing if I have time and try to setup a master->slave connection between the gps and path manager, as it was before, which will hopefully let me debug the issue.

PR #121, opened 2017-04-28
Kill mode implementation

sergei1152:
**Changes**
- Added StatusManager for managing heartbeat and uhf timers
- Added and tested kill mode implemenation. Its ugly right now but works


ianjfrosst:
This is kinda shitty. Oh well.

PR #130, opened 2017-08-10
altimeter calibrate

rosachen313:
-Change the value of offset to calibrate the altimeter while it's on the ground
-Pressure read by altimeter can be checked by switch the altimeter to barometer mode

### DIFF
--- a/Autopilot/Path Manager/MPL3115A2.c
+++ b/Autopilot/Path Manager/MPL3115A2.c
@@ -4,6 +4,7 @@
  *
  * Created on April 5, 2014, 7:08 PM
  */
+#include <stdio.h>
 #include "MPL3115A2.h"
 #include "main.h"
 #include "../Common/Utilities/Logger.h"
@@ -11,6 +12,8 @@
 float lastKnownAltitude = 0;
 char altimeterConnected = 0;
 float altimeterOffset = 0;
+char barometerConnected = 0;
+float lastKnownPressure = 0;
 
 char initAltimeter() {
     initI2C();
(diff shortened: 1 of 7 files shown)


Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -29,15 +32,33 @@ char initAltimeter() {
     return altimeterConnected;
 }
 
-void calibrateAltimeter(float altitude){
-    altimeterOffset = altitude;
+void calibrateAltimeter(long int pressure){
+    int msb = pressure >> 9;
```
sergei1152: these numbers look weird. Why a 9 bit shift?

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -29,15 +32,33 @@ char initAltimeter() {
...
+    int lsb = (pressure >> 1)& 255;
```
sergei1152: why a 1-bit shift here?

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -29,15 +32,33 @@ char initAltimeter() {
     return altimeterConnected;
 }
 
-void calibrateAltimeter(float altitude){
-    altimeterOffset = altitude;
+void calibrateAltimeter(long int pressure){
```
sergei1152: change the type to uint32_t

Autopilot/Path Manager/MPL3115A2.c
sergei1152: Change the type to a char or uint8_t

Autopilot/Path Manager/MPL3115A2.c
sergei1152: also do an and with a 0xFF instead

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -29,15 +32,33 @@ char initAltimeter() {
...
+    sendMessage(I2C_SLAVE_ADDRESS, BAR_IN_LSB, (char*)&lsb, 1, WRITE);
+    sendMessage(I2C_SLAVE_ADDRESS, OFF_H, (char*)&offset, 1, WRITE);
 }
 
 float getAltitude() {
     if (altimeterConnected && (sendMessage(I2C_SLAVE_ADDRESS, DATA_READY_REGISTER, 0, 0, READ) & 0x08)) {
-        int msb = sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_MSB_REGISTER, 0, 0, READ) << 8;
-        unsigned char csb = sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_CSB_REGISTER, 0, 0, READ);
-        float lsb = (sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_LSB_REGISTER, 0, 0, READ) >> 4)/16.0;
+        unsigned char pre_msb = sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_MSB_REGISTER, 0, 0, READ);
```
sergei1152: uint8_t should be the type instead

sergei1152 (changes requested):

Autopilot/Path Manager/MPL3115A2.c
rosachen313: That bit shift stuff is just what I read from the datasheet. Both altitude and pressure are 20 bit, but the way of reading those data are different.
screen shot 2017-08-12 at 7 55 05 pm

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -4,13 +4,16 @@
...
  */
+#include <stdio.h>
 #include "MPL3115A2.h"
 #include "main.h"
 #include "../Common/Utilities/Logger.h"
 
 float lastKnownAltitude = 0;
 char altimeterConnected = 0;
 float altimeterOffset = 0;
+char barometerConnected = 0;
```
sergei1152: use a bool instead. You'll have to import <stdbool.h>

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -30,36 +33,80 @@ char initAltimeter() {
...
 }
 
-void calibrateAltimeter(long int pressure){
-    int msb = pressure >> 9;
-    int lsb = (pressure >> 1)& 255;
-    int offset = 0;
-    sendMessage(I2C_SLAVE_ADDRESS, BAR_IN_MSB, (char*)&msb, 1, WRITE);
-    sendMessage(I2C_SLAVE_ADDRESS, BAR_IN_LSB, (char*)&lsb, 1, WRITE);
-    sendMessage(I2C_SLAVE_ADDRESS, OFF_H, (char*)&offset, 1, WRITE);
+void calibrateAltimeter(int altitudeOffset){
```
sergei1152: Now that i think of it, you can avoid having to do the error check if you just make the parameter type a `int8_t`, or an 8-bit signed integer

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -30,36 +33,80 @@ char initAltimeter() {
...
-        lastKnownAltitude = (float) (msb | csb) + lsb - altimeterOffset;
+        lastKnownAltitude = (float) (msb | csb) + lsb;
+ 
     }
     return lastKnownAltitude;
+}
+
+// toBarometer = 1 -> switch to Barometer
+// toBarometer = 0 -> switch to Altimeter
+void switchMode(int toBarometer) {
```
sergei1152: make this a static function (makes it only available to the local file). Same as making the function private in C++

Autopilot/Path Manager/main.h
```diff
@@ -14,7 +14,7 @@
 
 //Turns debug mode on or off. Usually contains small snippets of code to output
 //data through UART or to provide small input adjustments
-#define DEBUG 0
+#define DEBUG 1
```
sergei1152: disable this when putting up a production ready PR

Autopilot/Path Manager/PathManager.c
```diff
@@ -68,8 +68,13 @@ void pathManagerInit(void) {
     initInterchip(DMA_CHIP_ID_PATH_MANAGER);
 
     //Communication with Altimeter
+    
+    
     if (initAltimeter()){
-        calibrateAltimeter(GIVEN_PRESSURE);
+        debug("altimeter init");
```
sergei1152: no debug code should exist in PRs

Autopilot/Path Manager/MPL3115A2.h
```diff
@@ -29,15 +29,16 @@ extern "C" {
...
 
 
 
 //TODO: ADD BETTER FORMATTING TO THIS HEADER FILE
 char initAltimeter();
-void calibrateAltimeter(long int pressure);
+void calibrateAltimeter(int altitudeOffset);
 float getAltitude();
-
+void switchToBarometerMode(int toBarometer);
```
sergei1152: this function will never be used externally (only internally within the driver implementation). Therefore it should not exist in the .h file, and should only exist within the scope of the c file as a static function to keep it private.

Autopilot/Path Manager/MPL3115A2.c
sergei1152: a better way of doing this would be to declare an enum at the top and use that instead of this. Would make the code clearer. For example:
```c
typedef enum {
   ALTMETER,
   BAROMETER
} ALTMETER_MODE;
```
I think that syntax is right

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -30,36 +33,80 @@ char initAltimeter() {
...
+    if (altimeterConnected){
+        char data[3] = {0x38, 0x07, 0x39}; //switch to barometer mode
```
sergei1152: declare two defines  in the header file for the data instead. Eg.
```c
//in header file
#define SWITCH_TO_BAROMETER {0x38, 0x07, 0x39}
#define SWITCH_TO_ALTMETER { 0xB8, 0xB9, 0x39}

//then use it like this in your function:
char data[3] = SWITCH_TO_BAROMETER;

//or 
char data[3] = SWITCH_TO_ALTMETER;
```

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -30,36 +33,80 @@ char initAltimeter() {
...
+    if (altitudeOffset < -128 || altitudeOffset > 128){
+        error("Invalid altitude offset!");
+        // the range of altitude offset should be within -128 to 128
+    }
+    
+    else{
+        switchMode(1); //Barometer mode
+        float pressureFloat = 0.0;
+        //There's slight delay on pressure display, wait until pressure is not zero
+        while(pressureFloat <= 0.1){
```
sergei1152: You'll get a faster response time if you do a delay instead. Generally floating point operations on an embedded system like ours are extremely slow without a dedicated floating point unit, which the picpilot doesnt have. The delay can be up to 50us, which is quite alot. We have delay functions available for use here (look in Timer.h or Clock.h?) . You can experiment with different delays and choose the smallest one that gives you an accurate value (or see if its available in the datasheet.

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -30,36 +33,80 @@ char initAltimeter() {
...
-    sendMessage(I2C_SLAVE_ADDRESS, OFF_H, (char*)&offset, 1, WRITE);
+void calibrateAltimeter(int altitudeOffset){
+    if (altitudeOffset < -128 || altitudeOffset > 128){
+        error("Invalid altitude offset!");
+        // the range of altitude offset should be within -128 to 128
+    }
+    
+    else{
+        switchMode(1); //Barometer mode
+        float pressureFloat = 0.0;
```
sergei1152: you only want to do this pressure stuff if the input is 0 for the altmeter offset. Otherwise you should just adjust the altmeter offset register on the altmeter. You should also set the offset to 0 if you do happen to perform the calibration by setting the base pressure as 0

Autopilot/Path Manager/MPL3115A2.h
```diff
@@ -29,15 +29,16 @@ extern "C" {
 #define BAR_IN_LSB 0x15
 #define CONTROL_REGISTER1 0x26
 #define OFF_H 0x2D //offset register for alttitude
-#define GIVEN_PRESSURE 97913 //current pressure
+#define GIVEN_PRESSURE 97561 //current pressure
```
sergei1152: if you  arent using this, delete it

sergei1152 (changes requested):

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -27,18 +39,84 @@ char initAltimeter() {
...
-void calibrateAltimeter(float altitude){
-    altimeterOffset = altitude;
+
+void calibrateAltimeter(int8_t offsetInput){
+        // the range of altitude offset should be within -128 to 128
+    if (offsetInput == 0){
+        switchMode(BAROMETER); //Barometer mode
+        float pressureFloat = 0.0;
+        //There's slight delay on pressure display, wait until pressure is not zero
+        Delay_Us(370); //Delay time 370 ms
```
edenfrosst, line 52: You say delay 370 milliseconds, but you're using delay_us (microseconds).

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -27,18 +39,84 @@ char initAltimeter() {
...
     if (altimeterConnected && (sendMessage(I2C_SLAVE_ADDRESS, DATA_READY_REGISTER, 0, 0, READ) & 0x08)) {
-        int msb = sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_MSB_REGISTER, 0, 0, READ) << 8;
-        unsigned char csb = sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_CSB_REGISTER, 0, 0, READ);
-        float lsb = (sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_LSB_REGISTER, 0, 0, READ) >> 4)/16.0;
-        lastKnownAltitude = (float) (msb | csb) + lsb - altimeterOffset;
+        uint8_t pre_msb = sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_MSB_REGISTER, 0, 0, READ);
+        uint8_t pre_csb = sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_CSB_REGISTER, 0, 0, READ);
+        uint8_t pre_lsb = sendMessage(I2C_SLAVE_ADDRESS, ALTITUDE_LSB_REGISTER, 0, 0, READ);
+
+        int msb = pre_msb << 8;
```
edenfrosst, line 79: Use standard types for clarity

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -1,22 +1,34 @@
 /*
  * File:   MPL3115A2.c
- * Author: Chris Hajduk
+ * Author: Chris Hajduk & Rosa Chen
  *
  * Created on April 5, 2014, 7:08 PM
  */
+#include <stdio.h>
+#include <stdint.h>
+#include <stdlib.h>
```
edenfrosst, line 9: Most of these are included in main.h

Autopilot/Path Manager/MPL3115A2.c
```diff
@@ -1,22 +1,34 @@
...
 #include "MPL3115A2.h"
 #include "main.h"
 #include "../Common/Utilities/Logger.h"
+#include "../AttitudeManager/delay.h"
```
edenfrosst, line 13: These delay functions are old and inaccurate, and were removed for a reason. Better delay functions are available by including xc.h (which should be included in main.h).

edenfrosst (changes requested):
See elsewhere in the code for proper usage of delay functions.

PR #132, opened 2017-09-27
Bootcamp by January Huang

JanuaryHuang:
get the data and write codes for opening and closing the probes


Autopilot/AttitudeManager/Network/Commands.h
```diff
@@ -74,6 +74,7 @@ typedef const enum {
     SET_ALTITUDE_GAINS = 143,
     SET_GROUND_SPEED_GAINS = 144,
     CALIBRATE_PWM_INPUTS = 145,
+    SET_PROBE = 146,
```
edenfrosst: This section is for Multi-part commands (commands with more than one data value). This should be a value less than 128.

edenfrosst (changes requested):
Mostly good. Don't include the `.DS_Store` files with commits. They don't need to be under version control, and should be removed.

edenfrosst:
All good. I'll add you to the Asana.

